PR #5, opened 2017-11-24
Restructure for v0.9.1 to separate value from logic

### LINKED ISSUE #2: Separate value and logic on Value or Utility chain
- [x] separate logic and value on both chains (`SimpleStake.sol`, `STPrime.sol`, `BrandedToken.sol` are `ProtocolVersioned.sol` for `OpenSTValue.sol` and `OpenSTUtility.sol`
- [x] `proposeBrandedToken` and `registerBrandedToken onlyRegistrar` on Utility chain, and `registerUtilityToken onlyRegistrar` on value chain; name reservation on utility chain
- [x] re-implement staking

----
clean-up: solc does not allow import of relative paths

**problem** for readability we had moved the contracts that get deployed on the utility chain into `contracts/utilityChain/` and on the value chain into `contracts/valueChain`; but solc deliberately does not access other files on the filesystem without an explicit directory remapping (see https://github.com/ethereum/solidity/issues/757)

**solution** as the reorg was done for readability, we can opt for a prefix to the contracts filename and collapse the directory structure back to `contracts/`; contracts that need to be deployed only on the value chain will have a prefix `vc` as `vcMyContract.sol`, and contracts that only need to be deployed on the utility chain will have a prefix `uc`;  contracts that are deployed on both sides, or are inherited don't have a prefix.

benjaminbollen:
Refactor for v0.9.1: separate value staked into SimpleStake out of the versioned logic contract OpenSTValue; similarly BrandedToken.sol and STPrime.sol hold the utility tokens on the utility chain separate from the logic in OpenSTUtility.sol

Refactor is done up to staking, to be done for unstaking.

fixes #2


contracts/UtilityTokenAbstract.sol
```diff
@@ -0,0 +1,139 @@
...
+     */
+    function UtilityTokenAbstract(address _protocol, bytes32 _uuid)
+    	public
+    	ProtocolVersioned(_protocol)
+    {
+    	uuid = _uuid;
+    	totalTokenSupply = 0;
+    }
+
+	// /// @dev transfer full claim to beneficiary
```
jasonklein: Backslash mania

PR #7, opened 2017-11-24
ST' needs payable Initialize() to accept all base tokens

### LINKED ISSUE #6: Payable Initialize() for Simple Token Prime [ST']
On chain setup transfer the base tokens to STPrime.sol contract using `STPrime.Initialize()`

benjaminbollen:
hold off until #5 is merged

before the registrar registers a core on the value chain it must verify that the genesis exactly specified TOKENS_MAX so that all base tokens are held by STPrime.

added modifier `onlyInitialized` to `claim`, `mint`, `burn` on STPrime.sol

fixes #6

### DIFF
--- a/contracts/STPrime.sol
+++ b/contracts/STPrime.sol
@@ -43,6 +43,22 @@ import "./STPrimeConfig.sol";
 contract STPrime is UtilityTokenAbstract, STPrimeConfig {
 	using SafeMath for uint256;
 
+
+    /*
+     *  Storage
+     */
+    /// set when ST' has received TOKENS_MAX tokens;
+    /// when uninitialised minting is not allowed
+    bool private initialized;
+
+    /*
+     *  Modifiers
+     */
+    modifier onlyInitialized() {
+        require(initialized);
+        _;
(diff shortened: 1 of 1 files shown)


contracts/STPrime.sol
```diff
@@ -43,6 +43,22 @@ import "./STPrimeConfig.sol";
 contract STPrime is UtilityTokenAbstract, STPrimeConfig {
 	using SafeMath for uint256;
 
+
+    /*
+     *  Storage
+     */
+    /// set when ST' has received TOKENS_MAX tokens;
+    /// when uninitialised minting is not allowed
+    bool private initialized;
```
jasonklein: Why `private`?

benjaminbollen: because I felt it should not be exposed / considered for ST' to ever be un-initialised; if ST' is not yet initialised the utility chain is not even worth considering for adding as a Core

jasonklein: OK

jasonklein:
LGTM

Issue #21, opened 2017-11-29
Contracts: set min redemption/unstaking


jasonklein:
There should be a minimum amount required to process a redemption/unstaking (e.g., 1 UT and not 1 UT Wei) to facilitate conversion from UT to ST.

Issue #26, opened 2017-12-01
processStaking and processRedemption fallback


benjaminbollen:
for v0.9.1 we put a fallback to execute processStaking and processRedemption by Registrar in advance of separating the signature confirmation from the transaction formulation.

addresses #8 

for forcing cost
- [x] processStaking by Registrar
- [x] processRedemption by Registrar
- [x] unit test processStaking
- [x] unit test processRedemption

benjaminbollen:
@jasonklein this needs to applied for processRedemption still

PR #35, opened 2017-12-05
Unit Tests for Registrar and Protocol contracts

jasonklein:
This provides (additional) unit testing of `Registrar.sol`, `OpenSTValue.sol`, and `OpenSTUtility.sol`.

There are 3 failing tests. In `OpenSTValue.sol`, 2 tests are failing due to the math converting from BT to ST; see #21 . In `OpenSTUtility.sol`, there is one failing test due to a mistyped argument for a function. This test should pass after the merging of #33.


test/OpenSTUtility.js
```diff
@@ -78,6 +77,15 @@ const STPrime = artifacts.require("./STPrime.sol");
 /// 	fails to redeem when nonce is not > previously
 /// 	successfully redeems
 ///
+/// ProcessRedeeming
+///		BrandedToken
+/// 		fails to process if redemptionIntentHash is empty
+/// 		fails to process if msg.sender != redeemer
```
benjaminbollen: this will be changed, like I did for ProcessStaking with the fallback option;  Good for this PR; will then complete in #26 where we implement the fallback

PR #41, opened 2017-12-06
minimum redemption amount required to conversionRate UT Wei

### LINKED ISSUE #39: ConfirmRedemptionIntent must fail if redeeming less than 1 STWei
failing unit test

```
1) Contract: OpenSTValue ConfirmRedemptionIntent fails to confirm when amountUT does not convert into at least 1 STWei:
     AssertionError: Did not throw as expected
      at Object.module.exports.expectThrow (test/lib/utils.js:79:5)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

benjaminbollen:
for minimal precision of unstaking 1 ST Wei, we require minimum redemption amount to be required to be conversionRate UT Wei

fixes #39

### DIFF
--- a/contracts/OpenSTValue.sol
+++ b/contracts/OpenSTValue.sol
@@ -274,6 +274,8 @@ contract OpenSTValue is OpsManaged, Hasher {
     	expirationHeight = block.number + BLOCKS_TO_WAIT_SHORT;
 
     	UtilityToken storage utilityToken = utilityTokens[_uuid];
+    	// minimal precision to unstake 1 STWei
+    	require(_amountUT >= utilityToken.conversionRate);
     	amountST = _amountUT.div(utilityToken.conversionRate);
 
     	require(valueToken.balanceOf(address(utilityToken.simpleStake)) >= amountST);


jasonklein:
@benjaminbollen, I struck out my discussion above w/r/t rounding because I just came to where you came to the same conclusion elsewhere.

benjaminbollen:
Yes, @jasonklein current logic does not hold the `conversionRate` on the utilitychain side; this is a bit of an unfortunate UX; but we are still talking about negligible amounts; so negligible that it is hard to phantom how small;  I would rather put in a more reasonable minimum redemption amount related to the actual cost to unstake; so we can open an issue for that:

microtransactions in BT is not the same as unstaking micro-amounts; even unstaking O(1) BT will not be cost effective.

jasonklein:
@benjaminbollen, to confirm: (1) the solution is to place a requirement on the value chain that is not mirrored on the utility chain and (2) that solution is preferred over a comment about rounding/floating point math?

benjaminbollen:
Correct; also very explicitly lets discuss this as a new issue and move this forward; merge and move

PR #54, opened 2017-12-12
truffle: enable 200 optimizer runs on tests

### LINKED ISSUE #53: Run truffle test with optimised compilation
As it stands OpenSTValue and OpenSTUtility are quite heavy for a smart contract, and they hit 5M gas and 8M gas for deployment as unoptimised code;  To get a more realistic picture of actual gas consumption early-on in development, run truffle tests to also compile with 200 optimisation runs. 

This slows down `truffle test` but it additionally makes it more stable as it avoids cascading out of gas errors on the testRPC framework.

benjaminbollen:
fixes #53 

this now produces the exact same gas consumption on in the integration test of openst-protocol as it does on the deployment tests of openst-platform on ropsten:

```
OpenSTValue.new                            2889358
OpenSTUtility.new                            4503101
```

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -5,5 +5,14 @@ module.exports = {
       port: 8545,
       network_id: "*" // Match any network id
     }
+  },
+  solc: {
+    optimizer: {
+      enabled: true,
+      // set to same number of runs as openst-platform
+      // so that integration tests on openst-protocol
+      // give accurate gas measurements
+      runs: 200
+    }
   }
 };


truffle.js
```diff
@@ -5,5 +5,11 @@ module.exports = {
...
       network_id: "*" // Match any network id
     }
+  },
+  solc: {
+    optimizer: {
+      enabled: true,
+      runs: 200
+    }
   }
 };
```
abhayks1: should we add little comment on optimizer specifically on runs: 200?

benjaminbollen: will add

Issue #36, opened 2017-12-05
Control protocol versioning with Registrar


benjaminbollen:
in `ProtocolVersioned.sol` protocol transfer can be initiated by protocol.  call from Registrar through OpenST to initiate protocol transfer

PR #55, opened 2017-12-12
Admin initiate and revoke protocol transfer

benjaminbollen:
Enable a very simple for of protocol transfer to begin with on 0.9.1 OpenSTValue and OpenSTUtility to transfer protocol. Addresses #36

### DIFF
--- a/contracts/OpenSTUtility.sol
+++ b/contracts/OpenSTUtility.sol
@@ -31,6 +31,7 @@ import "./STPrime.sol";
 import "./STPrimeConfig.sol";
 import "./BrandedToken.sol"; 
 import "./UtilityTokenInterface.sol";
+import "./ProtocolVersioned.sol";
 
 
 /// @title OpenST Utility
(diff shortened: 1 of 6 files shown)


contracts/OpenSTUtility.sol
```diff
@@ -601,4 +601,28 @@ contract OpenSTUtility is Hasher, OpsManaged {
     		address(registeredToken.token),
     		registeredToken.registrar);
     }
+
+	/*
+	 *  Administrative functions
+	 */
+	function initiateProtocolTransfer(
+		// including ProtocolVersioned interface explicitly adds to the size
```
benjaminbollen: I'll need to change this comment

contracts/OpenSTUtility.sol
```diff
@@ -601,4 +601,28 @@ contract OpenSTUtility is Hasher, OpsManaged {
...
+		// of OpenSTValue, so we make it explicitly SimpleStake which is already included.
+		UtilityTokenAbstract _token,
+		address _proposedProtocol)
+		public
+		onlyAdmin
+		returns (bool)
+	{
+		_token.initiateProtocolTransfer(_proposedProtocol);
```
jasonklein: Need to explicitly return, otherwise will always return false. This comment applies to the identically-named function in `OpenSTValue.sol` as well.

contracts/OpenSTUtility.sol
```diff
@@ -601,4 +601,28 @@ contract OpenSTUtility is Hasher, OpsManaged {
...
+	}
+
+	function revokeProtocolTransfer(
+		UtilityTokenAbstract _token)
+		public
+		onlyAdmin
+		returns (bool)
+	{
+		_token.revokeProtocolTransfer();
```
jasonklein: Need to explicitly return, otherwise will always return false. This comment applies to the matching function in `OpenSTValue.sol` as well.

PR #66, opened 2017-12-14
Unit test case uncommented for complete protocol with mock contract f…

deepesh-kn:
…or protocol versioning

### DIFF
--- a/contracts/ProtocolVersionedMock.sol
+++ b/contracts/ProtocolVersionedMock.sol
@@ -1,0 +1,41 @@
+pragma solidity ^0.4.17;
+
+// Copyright 2017 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


contracts/ProtocolVersionedMock.sol
```diff
@@ -0,0 +1,41 @@
...
+    /*
+     *  Public functions
+     */
+    function ProtocolVersionedMock(address _protocol)
+        ProtocolVersioned(_protocol)
+        public { }
+
+    function blocksToWaitForProtocolTransfer() public pure returns (uint256) {
+        return PROTOCOL_TRANSFER_BLOCKS_TO_WAIT;
+    }
```
abhayks1: @jasonklein curious to know if blocksToWaitForProtocolTransfer is needed in ProtocolVersionedMock.sol?

This method is already there in ProtocolVersioned.sol. I guess solidity should inherit blocksToWaitForProtocolTransfer from ProtocolVersioned.sol to ProtocolVersionedMock.sol.

jasonklein: @abhayks1: sorry, I merged before I saw this question! Yes, you need to overwrite because 
`ProtocolVersioned.blocksToWaitForProtocolTransfer` points to the `PROTOCOL_TRANSFER_BLOCKS_TO_WAIT` constant scoped to it, and not scoped to `ProtocolVersionedMock`.

abhayks1: Thanks Banks for clarifying. Solidity is little different in method overriding as compared to other languages like Java, Ruby :)

PR #73, opened 2017-12-15
Import unit tests for Owned, OpsManaged and SafeMath

### LINKED ISSUE #4: Migrate tests from SimpleTokenSale
Certain contracts used in SimpleTokenSale are also used in this repository. Where there are already tests for those contracts, they should be included in this repository and updated as needed. Those include:

- [x] OpsManaged.sol
- [x] Owned.sol
- [x] SafeMath.sol

benjaminbollen:
fixes #4

### DIFF
--- a/contracts/SafeMathMock.sol
+++ b/contracts/SafeMathMock.sol
@@ -1,0 +1,45 @@
+pragma solidity ^0.4.17;
+
+// Copyright 2017 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


contracts/SafeMathMock.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity ^0.4.17;
+
+// Copyright 2017 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
```
abhayks1: any specific reason for SafeMatchMock contract? Can't we use SafeMatch contract in SafeMath.js directly?

contracts/SafeMathMock.sol
```diff
@@ -0,0 +1,45 @@
...
+// Based on the SafeMath library by the OpenZeppelin team.
+// Copyright (c) 2016 Smart Contract Solutions, Inc.
+// https://github.com/OpenZeppelin/zeppelin-solidity
+// The MIT License.
+// ----------------------------------------------------------------------------
+
+import "./SafeMath.sol";
+
+
+contract SafeMathMock {
```
abhayks1: Do we need "is SafeMath" to inherit SafeMath.sol functionality?

benjaminbollen: the idea is to have this contract call on SafeMath.sol (and inheriting it may tell the compiler to optimise it out)

Issue #45, opened 2017-12-07
Unit tests: reversion functions


jasonklein:
Add unit tests for:

- [x] OpenSTValue.revertStaking
- [x] OpenSTValue.revertUnstaking
- [x] OpenSTUtility.revertMinting
- [x] OpenSTUtility.revertRedemption

abhayks1:
#68

kushal-ladha:
Revert Minting: #67

deepesh-kn:
OpenSTValue.revertStaking: #69

PR #74, opened 2017-12-19
Unit test Cases for  Revert Staking, Minting, Revert Redemption, Revert Unstaking

abhayks1:


contracts/OpenSTValue.sol
```diff
@@ -380,20 +380,20 @@ contract OpenSTValue is OpsManaged, Hasher {
 		
 		Unstake storage unstake = unstakes[_redemptionIntentHash];
 
-    	// require that the unstake has expired and that the redeemer has not
-    	// processed the unstaking, ie unstake has not been deleted
-    	require(unstake.expirationHeight > 0);
-    	require(unstake.expirationHeight <= block.number);
+			// require that the unstake has expired and that the redeemer has not
```
benjaminbollen: we will need to discuss a standard and then implement a Linter check in the CI (to be put in place); for now OK, but the editors fighting is really not clean on the origin repo

test/OpenSTUtility.js
```diff
@@ -535,6 +536,144 @@ contract('OpenSTUtility', function(accounts) {
...
-})
+
+	// Unit test cases for revert rdemption
+	describe('revert redemption', async () => {
+
+		context('revert redemption', async () => {
+
+			var stakeAmountST = 1,
+					convertedAmountBT = (stakeAmountST*conversionRate),
+					redemptionAmountBT = 2, // How many Branded tokens to redeem
```
benjaminbollen: This is in BT Wei, not BT; please correct to BT

contracts/OpenSTValue.sol
kushal-ladha: Hey Ben, me, deepesh and abhay have synched are tab space to two. We can decide on which linter to use as we all use different editors for coding.

test/OpenSTValue.js
```diff
@@ -506,4 +508,250 @@ contract('OpenSTValue', function(accounts) {
...
+
+
+	// Revert Staking before ProcessStaking
+	describe('Process RevertStaking before ProcessStaking', async () => {
+
+		context('Revert Stake', async () => {
+
+			var staker = accounts[0];			
+			var owner  = accounts[1];
+			var amountST = 126;
```
benjaminbollen: We should do full scan of unit tests of consistent use of ST vs STWei (UT vs UTWei, BT vs BTWei)

benjaminbollen:
closes #45 
closes #67 
closes #68 
closes #69

Issue #92, opened 2018-01-18
STPrimePayable - remove


jasonklein:
Remove STPrimePayable.sol as no longer necessary.

PR #93, opened 2018-01-18
Clean-up: removals

jasonklein:
Removes files that are no longer used or usable:
- STPrimePayable.sol
- compile.sh

Also some tidying changes to `tests/lib/utils.js`.


benjaminbollen:
closes #92

PR #110, opened 2018-02-20
conversion rate be a (bound) rational number

deepesh-kn:
This includes the following issue: 
https://github.com/OpenSTFoundation/openst-protocol/issues/107

### DIFF
--- a/contracts/BrandedToken.sol
+++ b/contracts/BrandedToken.sol
@@ -53,7 +53,7 @@ contract BrandedToken is EIP20Token, UtilityTokenAbstract {
         uint256 _chainIdValue,
         uint256 _chainIdUtility,
         uint256 _conversionRate,
-        uint256 _conversionRateDecimalFactor)
+        uint8 _conversionRateDecimalFactor)
         public
         EIP20Token(_symbol, _name, _decimals)
         UtilityTokenAbstract(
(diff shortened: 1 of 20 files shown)


contracts/BrandedToken.sol
```diff
@@ -53,7 +53,7 @@ contract BrandedToken is EIP20Token, UtilityTokenAbstract {
         uint256 _chainIdValue,
         uint256 _chainIdUtility,
         uint256 _conversionRate,
-        uint256 _conversionRateDecimalFactor)
+        uint8 _conversionRateDecimalFactor)
```
benjaminbollen: now it's not the factor anymore and perhaps can be renamed `_conversionRateDecimals`

contracts/STPrimeConfig.sol
```diff
@@ -29,6 +29,7 @@ contract STPrimeConfig {
     string  public constant STPRIME_NAME            = "SimpleTokenPrime";
     uint256 public constant STPRIME_CONVERSION_RATE = 1;
     uint8   public constant TOKEN_DECIMALS          = 18;
+    uint8   public constant STPRIME_CONVERSION_RATE_TOKEN_DECIMALS = 5;
```
benjaminbollen: this constant can be 1

benjaminbollen: correction: it can be 0

contracts/BrandedToken.sol
deepesh-kn: Done

contracts/STPrimeConfig.sol
deepesh-kn: Done, changed to 0.

PR #125, opened 2018-05-02
Add gate contract 

noslav:
Adds gate contract
* gate contract is ProtocolVersioned
* declares a struct StakeRequest
* defines a mapping between staker address and StakeRequest 
* declares state variable workers

### DIFF
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -1,0 +1,46 @@
+pragma solidity ^0.4.23;
+
+// Copyright 2018 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


contracts/Gate.sol
```diff
@@ -0,0 +1,46 @@
...
+import "./ProtocolVersioned.sol";
+
+contract Gate is ProtocolVersioned {
+    /*
+     *  Structures
+     */
+    struct StakeRequest {
+        uint256 amount;
+        uint256 unlockHeight;
+        address beneficiary; //The token holder contract in the future
```
benjaminbollen: the value of the beneficiary might be enforced to be a token holder contract, but the concept of the `beneficiary` remains

PR #127, opened 2018-05-04
Gate contract: requestStake, revertStakeRequest and rejectStakeRequest implementation 

deepesh-kn:
This includes: 
1. requestStake implementation
2. revertStakeRequest implementation
3. rejectRequest implementation.

### DIFF
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -75,7 +75,7 @@ contract Gate is ProtocolVersioned, Owned {
     {
         require(_workers != address(0));
         require(_openSTValue != address(0));
-        require(_uuid.length != 0);
+        require(_uuid.length != bytes32(0));
 
         workers = _workers;
         bounty = _bounty;
@@ -90,7 +90,7 @@ contract Gate is ProtocolVersioned, Owned {
         returns ( bool /* success */)
     {
 
-        require(_amount > 0);
+        require(_amount > uint256(0));
         require(_beneficiary != address(0));
 
         // check if the stake request does not exists
@@ -123,7 +123,7 @@ contract Gate is ProtocolVersioned, Owned {
         require(stakeRequest.beneficiary != address(0));
 
         // check if the stake request was not accepted
-        require(stakeRequest.hashLock == 0);
+        require(stakeRequest.hashLock == bytes32(0));
 
         require(OpenSTValueInterface(openSTProtocol).valueToken().transfer(msg.sender, stakeRequest.amount));
 
@@ -148,7 +148,7 @@ contract Gate is ProtocolVersioned, Owned {
         require(stakeRequest.beneficiary != address(0));
 
         // check if the stake request was not accepted
-        require(stakeRequest.hashLock == 0);
+        require(stakeRequest.hashLock == bytes32(0));
 
         // transfer the amount back
         require(OpenSTValueInterface(openSTProtocol).valueToken().transfer(_staker, stakeRequest.amount));
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -97,8 +97,8 @@ contract Gate is ProtocolVersioned {
         require(_amount > 0);
         require(_beneficiary != address(0));
 
-        // check if the state request does not exists
-        require(address(stakeRequests[msg.sender].beneficiary) == address(0));
+        // check if the stake request does not exists
+        require(stakeRequests[msg.sender].beneficiary == address(0));
 
         require(openSTValue.valueToken().allowance(msg.sender, address(this)) >= _amount);
         require(openSTValue.valueToken().transferFrom(msg.sender, address(this), _amount));
@@ -121,8 +121,8 @@ contract Gate is ProtocolVersioned {
     {
         StakeRequest storage stakeRequest = stakeRequests[msg.sender];
 
-        // check if the state request exists for the msg.sender
-        require(address(stakeRequest.beneficiary) != address(0));
+        // check if the stake request exists for the msg.sender
+        require(stakeRequest.beneficiary != address(0));
 
         // check if the stake request was not accepted
         require(stakeRequest.unlockHeight == 0);
@@ -143,19 +143,18 @@ contract Gate is ProtocolVersioned {
         // check if the caller is whitelisted worker
         //require(workers.isWorker(msg.sender)); //TODO: revist this to add worker check
 
-        // check if the stake request was done.
         StakeRequest storage stakeRequest = stakeRequests[_staker];
 
-        // check if the state request exists
-        require(address(stakeRequest.beneficiary) != address(0));
+        // check if the stake request exists
+        require(stakeRequest.beneficiary != address(0));
 
         // check if the stake request was not accepted
         require(stakeRequest.unlockHeight == 0);
 
         // transfer the amount back
-        require(openSTValue.valueToken().transfer(msg.sender, stakeRequest.amount));
+        require(openSTValue.valueToken().transfer(_staker, stakeRequest.amount));
 
-        // delete the state request from the mapping storage
+        // delete the stake request from the mapping storage
         delete stakeRequests[msg.sender];
 
         emit StakeRequestRejected(_staker, stakeRequest.amount);
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -75,7 +75,7 @@ contract Gate is ProtocolVersioned, Owned {
     {
         require(_workers != address(0));
         require(_openSTValue != address(0));
-        require(_uuid.length != bytes32(0));
+        require(_uuid.length != uint8(0));
 
         workers = _workers;
         bounty = _bounty;
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -109,7 +109,7 @@ contract Gate is ProtocolVersioned {
             unlockHeight: 0
             });
 
-        StakeRequested(msg.sender, _amount, _beneficiary, unlockHeight);
+        StakeRequested(msg.sender, _amount, _beneficiary, 0);
 
         return unlockHeight;
     }
@@ -139,7 +139,7 @@ contract Gate is ProtocolVersioned {
         returns (bool)
     {
         // check if the caller is whitelisted worker
-        require(workers.isWorker(msg.sender));
+        require(workers.isWorker(msg.sender)); //TODO: revist this
 
         // check if the stake request was done.
         StakeRequest storage stakeRequest = stakeRequests[_staker];
@@ -147,6 +147,9 @@ contract Gate is ProtocolVersioned {
         // check if the state request exists
         require(address(stakeRequest.beneficiary) != address(0));
 
+        // check if the stake request was not accepted
+        require(stakeRequest.unlockHeight == 0);
+        
         // transfer the amount back
         require(EIP20Interface(openSTValue.valueToken).transfer(msg.sender, stakeRequest.amount));
 
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.4.23;
+pragma solidity ^0.4.23;
 
 // Copyright 2018 OpenST Ltd.
 //
@@ -60,7 +60,7 @@ contract Gate is ProtocolVersioned {
     bytes32 uuid;
 
     // OpenSTValue contract
-    address openSTValue;
+    OpenSTValue openSTValue;
 
     /*
      *  Public functions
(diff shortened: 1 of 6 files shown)


contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
+        bytes32 _uuid,
+        address _openSTValue)
+        public
+        ProtocolVersioned(_openSTValue)
+    {
+        require(_workers != address(0));
+        require(_openSTValue != address(0));
+        require(_uuid.length != 0);
+
+        // Should bounty with 0 amount be allowed ?
```
jasonklein: For the purposes of limiting the PoC scope, these kinds of questions we will set aside. Our testing will focus on the complete paths and not validations, so no need for validation `require`s.

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
     }
+
+    function rejectStakeRequest(address _staker)
+        external
+        returns (bool /* success */)
+    {
+        // check if the caller is whitelisted worker
+        //require(workers.isWorker(msg.sender)); //TODO: revist this to add worker check
+
+        // check if the stake request was done.
```
noslav: this comment is confusing.

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
+        StakeRequest storage stakeRequest = stakeRequests[_staker];
+
+        // check if the state request exists
+        require(address(stakeRequest.beneficiary) != address(0));
+
+        // check if the stake request was not accepted
+        require(stakeRequest.unlockHeight == 0);
+
+        // transfer the amount back
+        require(openSTValue.valueToken().transfer(msg.sender, stakeRequest.amount));
```
noslav: the amount shouldn't be going to the facilitator but rather the staker.

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
+        require(address(stakeRequest.beneficiary) != address(0));
+
+        // check if the stake request was not accepted
+        require(stakeRequest.unlockHeight == 0);
+
+        require(openSTValue.valueToken().transfer(msg.sender, stakeRequest.amount));
+
+        delete stakeRequests[msg.sender];
+
+        emit StakeRequestReverted(msg.sender, stakeRequest.amount);
```
noslav: would stakeRequest.amount would be zero after its deleted from the mapping? This question also applies to the other delete in the reject function.

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
      *  Storage
      */
     address public workers; //TODO: update type once workers contract and interfaces are added to the repo.
+
+    // stake requests
     mapping(address /*staker */ => StakeRequest) public stakeRequests;
 
-    function Gate() public { //TODO: complete as necessary
+    // bounty amount
+    uint256 private bounty;
```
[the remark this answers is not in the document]
deepesh-kn: As of now Iam not sure if the bounty will be accessed outside the contract. So keeping it as private as of now. When we implement bounty related code we can check what should be its accessibility and change accordingly.

contracts/Gate.sol
deepesh-kn: sure!!!

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
+    // utility token UUID
+    bytes32 private uuid;
+
+    // OpenSTValue contract
+    OpenSTValue private openSTValue;
+
+    /*
+     *  Public functions
+     */
+    constructor(
```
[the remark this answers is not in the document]
deepesh-kn: If we use ``` function Gate() public ``` we will get deprecation warning, for compiler using 0.4.23. 
```
Warning: Defining constructors as functions with the same name as the contract is deprecated. Use "constructor(...) { ... }" instead.

```

contracts/Gate.sol
```diff
@@ -38,9 +50,117 @@ contract Gate is ProtocolVersioned {
...
+        address _beneficiary)
+        external
+        returns ( bool /* success */)
+    {
+
+        require(_amount > 0);
+        require(_beneficiary != address(0));
+
+        // check if the state request does not exists
+        require(address(stakeRequests[msg.sender].beneficiary) == address(0));
```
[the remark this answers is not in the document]
deepesh-kn: Removed it.

contracts/Gate.sol
deepesh-kn: Removed the comment, It was not supposed to be there!!!

contracts/Gate.sol
deepesh-kn: Good catch !!!

contracts/Gate.sol
deepesh-kn: Yes, stakeRequest.amount would be 0

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+        external
+        returns (bool /* success */)
+    {
+        StakeRequest storage stakeRequest = stakeRequests[msg.sender];
+
+        // check if the stake request exists for the msg.sender
+        require(stakeRequest.beneficiary != address(0));
+
+        // check if the stake request was not accepted
+        require(stakeRequest.unlockHeight == 0);
```
benjaminbollen: let's define the state of "stake-exists-and-no-HTL" that `beneficiary != 0x0 & hashLock = 0x0`

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+        // check if the caller is whitelisted worker
+        //require(workers.isWorker(msg.sender)); //TODO: revist this to add worker check
+
+        StakeRequest storage stakeRequest = stakeRequests[_staker];
+
+        // check if the stake request exists
+        require(stakeRequest.beneficiary != address(0));
+
+        // check if the stake request was not accepted
+        require(stakeRequest.unlockHeight == 0);
```
benjaminbollen: similar correction for the definition of the state, let's use `hashLock != 0` rather than `unlockHeight`

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+
+        // transfer the amount back
+        require(openSTValue.valueToken().transfer(_staker, stakeRequest.amount));
+
+        // delete the stake request from the mapping storage
+        delete stakeRequests[msg.sender];
+
+        emit StakeRequestRejected(_staker, stakeRequest.amount);
```
benjaminbollen: similar question concerning the value of `stakeRequest.amount` after deletion of storage variable on L158

contracts/Gate.sol
deepesh-kn: Ahh, I overlooked the comment before. Yes it should return the actual value. Made changes

contracts/Gate.sol
```diff
@@ -22,8 +23,19 @@ pragma solidity ^0.4.23;
 // ----------------------------------------------------------------------------
 
 import "./ProtocolVersioned.sol";
+import "./OpenSTValue.sol";
+import "./EIP20Interface.sol";
+
 
 contract Gate is ProtocolVersioned {
```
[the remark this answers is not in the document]
deepesh-kn: Done

contracts/Gate.sol
```diff
@@ -22,8 +23,19 @@ pragma solidity ^0.4.23;
 // ----------------------------------------------------------------------------
 
 import "./ProtocolVersioned.sol";
+import "./OpenSTValue.sol";
```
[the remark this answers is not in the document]
deepesh-kn: cannot use ```import {OpenSTValueInterface} from "./OpenSTValue.sol"``` as OpenSTValue is not implementing (inheriting) OpenSTValueInterface.

Using ``` import "./EIP20Interface.sol"; ```
Due to this I had to declare ``` valueToken() ``` in OpenSTValueInterface.

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
 
-    function Gate() public { //TODO: complete as necessary
+    // bounty amount
+    uint256 private bounty;
+
+    // utility token UUID
+    bytes32 private uuid;
+
+    // OpenSTValue contract
+    OpenSTValue private openSTValue;
```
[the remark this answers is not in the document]
deepesh-kn: Removed this and using ``` openSTProtocol ``` instead

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+        require(_openSTValue != address(0));
+        require(_uuid.length != 0);
+
+        // Should bounty with 0 amount be allowed ?
+        //require(_bounty > 0);
+
+        workers = _workers;
+        bounty = _bounty;
+        uuid = _uuid;
+        openSTValue = OpenSTValue(_openSTValue);
```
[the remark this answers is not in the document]
deepesh-kn: Removed

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+        returns ( bool /* success */)
+    {
+
+        require(_amount > 0);
+        require(_beneficiary != address(0));
+
+        // check if the stake request does not exists
+        require(stakeRequests[msg.sender].beneficiary == address(0));
+
+        require(openSTValue.valueToken().allowance(msg.sender, address(this)) >= _amount);
```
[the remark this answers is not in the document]
deepesh-kn: removed

contracts/Gate.sol
```diff
@@ -38,9 +50,116 @@ contract Gate is ProtocolVersioned {
...
+        emit StakeRequested(msg.sender, _amount, _beneficiary);
+
+        return true;
+    }
+
+    function revertStakeRequest()
+        external
+        returns (bool /* success */)
+    {
+        StakeRequest storage stakeRequest = stakeRequests[msg.sender];
```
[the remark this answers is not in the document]
deepesh-kn: added

contracts/Gate.sol
deepesh-kn: changed to ``` hashLock = 0x0 ``` instead of unlockHeight

contracts/Gate.sol
deepesh-kn: changed to ``` hashLock = 0x0 ``` instead of unlockHeight

contracts/Gate.sol
deepesh-kn: Yes made the change

contracts/Gate.sol
```diff
@@ -100,8 +96,7 @@ contract Gate is ProtocolVersioned {
         // check if the stake request does not exists
         require(stakeRequests[msg.sender].beneficiary == address(0));
 
-        require(openSTValue.valueToken().allowance(msg.sender, address(this)) >= _amount);
-        require(openSTValue.valueToken().transferFrom(msg.sender, address(this), _amount));
+        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), _amount));
```
benjaminbollen: nice :)

contracts/Gate.sol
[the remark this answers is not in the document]
benjaminbollen: thx :)

contracts/Gate.sol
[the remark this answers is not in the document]
jasonklein: Oh, sorry! Thanks for catching that!

contracts/Gate.sol
```diff
@@ -22,8 +23,19 @@ pragma solidity ^0.4.23;
...
+
 
 contract Gate is ProtocolVersioned {
+
+    /*
+     * Events
+     */
+    event StakeRequested(address _staker, uint256 _amount, address _beneficiary);
+    event StakeRequestReverted(address _staker, uint256 _amount);
+    event StakeRequestRejected(address _staker, uint256 _amount);
```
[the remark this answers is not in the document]
jasonklein: You mean include a `reason` parameter in the events, @benjaminbollen?

contracts/Gate.sol
```diff
@@ -115,23 +110,27 @@ contract Gate is ProtocolVersioned {
...
     {
+        // only staker can do revertStakeRequest, msg.sender == staker
         StakeRequest storage stakeRequest = stakeRequests[msg.sender];
 
         // check if the stake request exists for the msg.sender
         require(stakeRequest.beneficiary != address(0));
 
         // check if the stake request was not accepted
-        require(stakeRequest.unlockHeight == 0);
+        require(stakeRequest.hashLock == 0);
```
benjaminbollen: can we use the constructor `bytes32(0)`, similarly as we use `address(0)`

contracts/Gate.sol
```diff
@@ -149,15 +148,16 @@ contract Gate is ProtocolVersioned {
...
```
benjaminbollen: also here `bytes32(0)`

PR #128, opened 2018-05-09
[Dont Merge]

deepesh-kn:


benjaminbollen:
1.a. OpenSTValue needs to clean up the mapping itself; I have to understand better where this requirement seemingly showed up. -- will get back to this
1.b. it is intentional to not have the double-check as the value held is controlled by OpenSTValue, not the Gate contract; and the whole transaction would hit REVERT if OpenSTValue fails it;  when OpenSTValue is turned into a library call, this requirement needs to move into the Gate contract (v0.9.4)
2.a similarly fine, as with 1.b.
2.b. on `processStaking` the bounty must return to the `workers` contract address, if msg.sender is a worker in that contract; if it not a worker, then the bounty must be returned to the msg.sender of `processStaking`; on `revertStaking`, the bounty must always go to the workers contract address of the gate.
3. -- im writing this in the car -- but from my understanding Gate contract does need to store unlockHeight  (perhaps not hashLock, but we may for 0.9.3 as per note 1.b knowing v0.9.4)

benjaminbollen:
I will review the code changes later today; sorry for the delay

contracts/Gate.sol
```diff
@@ -162,4 +169,130 @@ contract Gate is ProtocolVersioned, Owned {
...
+    require(stakeRequest.hashLock != bytes32(0));
+
+    // we call processStaking for OpenSTValue and get the stakeAddress on success.
+    address stakerAddress = OpenSTValueInterface(openSTProtocol).processStaking(_stakingIntentHash, _unlockSecret);
+
+    // check if the stake address is not 0
+    require(stakerAddress != address(0));
+
+    // Transfer bounty amount to the msg.sender account
+    require(OpenSTValueInterface(openSTProtocol).valueToken().transfer(msg.sender, bounty));
```
benjaminbollen: I've mentioned this in the general comment, but here we want to check that if `isWorker(msg.sender)` then transfer it to `workers`, the contract address, otherwise transfer to `msg.sender` (so as the Workers check is not yet enforced, this line is consistent, but best to already include the condition)

test/Gate.js
```diff
@@ -0,0 +1,294 @@
...
+    // check balances
+    assert.equal(finalStakerAccountBalance.equals(initialStakerAccountBalance.sub(stakeAmount)), true);
+    assert.equal(finalGateBalance.equals(initialGateBalance.plus(stakeAmount)), true);
+
+    // request again should fail
+    await Utils.expectThrow(gate.requestStake(stakeAmount, beneficiaryAccount, {from: stakerAccount}));
+  };
+
+  const acceptStakeRequest = async function() {
+    await valueToken.transfer(workerAddress1, new BigNumber(web3.toWei(10000, "ether")),{from: accounts[0]});
```
benjaminbollen: accounts[7] also has large balance for testing purposes; seems not needed?

contracts/Gate.sol
```diff
@@ -35,14 +35,22 @@ contract Gate is ProtocolVersioned, Owned {
...
+
 
     /*
      *  Structures
      */
     struct StakeRequest {
         uint256 amount;
         uint256 unlockHeight;
-        address beneficiary; //The token holder contract in the future 
+        address beneficiary; //The token holder contract in the future
```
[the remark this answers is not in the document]
deepesh-kn: Removed the comment !!!

contracts/Gate.sol
```diff
@@ -162,4 +169,130 @@ contract Gate is ProtocolVersioned, Owned {
...
+        require(_hashLock != bytes32(0));
+
+        // Transfer bounty amount form worker to gate contract
+        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), bounty));
+
+        // Approve OpenSTValue contract for stake amount
+        require(OpenSTValueInterface(openSTProtocol).valueToken().approve(openSTProtocol, stakeRequest.amount));
+
+
+        uint256 amountUT = 0;
```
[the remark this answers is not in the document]
deepesh-kn: Defined the variables in the returns in the function signature.

contracts/Gate.sol
```diff
@@ -162,4 +169,130 @@ contract Gate is ProtocolVersioned, Owned {
...
+
+        // Check if the stake function call did not result in to error.
+        require(stakingIntentHash != bytes32(0));
+
+        stakeRequests[_staker].unlockHeight = unlockHeight;
+        stakeRequests[_staker].hashLock = _hashLock;
+
+        emit StakeRequestAccepted(_staker, stakeRequest.amount, amountUT, nonce, unlockHeight, stakingIntentHash);
+
+        return true;
```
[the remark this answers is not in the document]
deepesh-kn: returning the 4 params now !!!

contracts/Gate.sol
```diff
@@ -162,4 +169,130 @@ contract Gate is ProtocolVersioned, Owned {
...
+    bytes32 _unlockSecret)
+    external
+    returns (bool /* success */)
+  {
+    // check if the caller is whitelisted worker
+    //require(workers.isWorker(msg.sender)); //TODO: revist this to add worker check
+
+    require(_stakingIntentHash != bytes32(0));
+
+    address staker = OpenSTValueInterface(openSTProtocol).getStakerAddress(_stakingIntentHash);
```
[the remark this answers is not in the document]
deepesh-kn: Added the comment !!!

contracts/Gate.sol
deepesh-kn: Added the commented code with TODO.

test/Gate.js
deepesh-kn: When I call ``` await valueToken.balanceOf.call(workerAddress1) ```, I get balance as 0. 
I checked all the account balance, only account[0] had the large balance. So I used transfer.

contracts/Gate.sol
```diff
@@ -216,7 +215,7 @@ contract Gate is ProtocolVersioned, Owned {
 
         emit StakeRequestAccepted(_staker, stakeRequest.amount, amountUT, nonce, unlockHeight, stakingIntentHash);
 
-        return true;
+        return (amountUT, nonce, unlockHeight, stakingIntentHash);
```
benjaminbollen: great! thanks!

test/Gate.js
```diff
@@ -78,12 +78,16 @@ contract('Gate', function(accounts) {
     await valueToken.approve(gate.address, bountyAmount, { from: workerAddress1 });
 
     let acceptStakeRequestResult = await gate.acceptStakeRequest.call(stakerAccount, lock.l, {from: workerAddress1});
-    assert.equal(acceptStakeRequestResult, true);
+    let amountUT = acceptStakeRequestResult[0];
+    let nonce = acceptStakeRequestResult[1];
+    let unlockHeight = acceptStakeRequestResult[2].plus(1);
+
+    let stakingIntentHash = await openSTValue.hashStakingIntent.call(uuid, stakerAccount, nonce, beneficiaryAccount, stakeAmount,
```
benjaminbollen: sweet! thx

test/Gate_utils.js
```diff
@@ -120,13 +120,27 @@ module.exports.checkStakeRequestRejectedEvent = (event, _staker, _amount) => {
   assert.equal(event.args._amount.toNumber(10), _amount.toNumber(10));
 };
 
-module.exports.checkStakeRequestAcceptedEvent = (event, _staker, _amount) => {
-  if (Number.isInteger(_amount)) {
-    _amount = new BigNumber(_amount);
+
+module.exports.checkStakeRequestAcceptedEvent = (event, _staker, _amountST, _amountUT, _nonce, _unlockHeight, _stakingIntentHash) => {
```
benjaminbollen: thanks! good catch to add this then

benjaminbollen:
the tests are failing though because of the changed signatures perhaps.

PR #130, opened 2018-05-09
Adds the workers contract

noslav:
* Adds the workers contract from the [openst-payments](https://github.com/OpenSTFoundation/openst-payments) repository.
* Sets an EIP20 token address in the contract constructor.
* Adds the functionality for calling the approve function for an address on an EIP20 token.
MOSA-27

### DIFF
--- a/contracts/Workers.sol
+++ b/contracts/Workers.sol
@@ -59,7 +59,7 @@ contract Workers is OpsManaged {
     {
         require(_eip20token != address(0));
         
-        eip20token = _eip20token;
+        eip20token = EIP20Interface(_eip20token);
     }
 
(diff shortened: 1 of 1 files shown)


contracts/Workers.sol
```diff
@@ -0,0 +1,146 @@
...
+        uint256 indexed _deactivationHeight,
+        uint256 _remainingHeight);
+
+    ///Event for worker removed
+    event WorkerRemoved(
+        address indexed _worker,
+        bool _existed);
+
+    /// @dev    Constructor;
+    ///         public method;
```
benjaminbollen: Update the comment with the param address eip20token, and what it is used for

PR #131, opened 2018-05-10
[MOSA-48]Upgraded web3 version to v1.0.0.beta.33

sarvesh-ost:


sarvesh-ost:
1. Upgraded web3 version to ^1.0.0-beta.33
2. Removed event.js dependency from the test as it's no longer supported. require("web3/lib/web3/event.js");
3. Updated test case in test/SimpleToken.js to use abi decoder.
4. Removed previous event decoder dependent on event.js dependency. 
5. Added one more method in EIP20Token_utils.js to assert transfer event decoded by abi decoder.

test/SimpleToken.js
```diff
@@ -101,377 +102,377 @@ const SimpleToken = artifacts.require("./SimpleToken/SimpleToken.sol")
...
-      })
-
-      it("totalSupply", async () => {
-         assert.equal((await token.totalSupply.call()).toNumber(), TOTAL_SUPPLY.toNumber())
-      })
 
-      it("balances is private", async () => {
-         assert.isTrue(typeof(token.balances) == 'undefined')
-      })
+  const SYMBOL = "ST"
```
sarvesh-ost: @benjaminbollen  @jasonklein @abhayks1 
SimpleToken.js was at 4 space indentation, however, we are following 2-2-2 space indentation. Due to this, the diff is looking larger than expected. 
I hope this is fine with you.

PR #132, opened 2018-05-14
Sarvesh/fixing compiler warnings

sarvesh-ost:


sarvesh-ost:
Syntax corrections, based on latest compiler version

1. Compiler statement
2. Change in emit event

PR #138, opened 2018-05-23
Sarvesh/mosa 44 verify proof

sarvesh-ost:
Verification of account proof using contract on production geth data. 

**Setup required to run test**
1. Sync production utility chain using geth
2. Copy chain data folder and specify path in test
3. Run geth(production instance) on local port 9545


contracts/mosaic/MerklePatriciaProof.sol
```diff
@@ -1,15 +1,11 @@
-/*
```
benjaminbollen: Let's keep the original author versioning intact; we can reference that it is based on, or derived from ... (with a link)

benjaminbollen: also, I believe these files were MIT, so we should keep such derived files explicitly MIT licensed in the header

contracts/mosaic/MerklePatriciaProof.sol
```diff
@@ -18,127 +14,79 @@ library MerklePatriciaProof {
...
      */
-    function verify(bytes32 value, bytes encodedPath, bytes rlpParentNodes, bytes32 root) internal pure returns (bool) {
+    function verify(bytes32 value, bytes addr, bytes rlpParentNodes, bytes32 root) public returns (bool) {
+
+        //Convert back RLP encoded parent Nodes to list of Nodes
         RLP.RLPItem memory item = RLP.toRLPItem(rlpParentNodes);
         RLP.RLPItem[] memory parentNodes = RLP.toList(item);
 
+        //Sha3 of account address - used from traversing from root to desired node
+        bytes memory encodedPath = bytes32ToBytes(keccak256(addr));
```
benjaminbollen: by providing the encoded path in the parameters the `verify` function is general for either an account lookup (which is stored under the sha3 of its address indeed) as well as for other values (ie for storage values inside the accounts storage)

contracts/mosaic/MerklePatriciaProof.sol
```diff
@@ -18,127 +14,79 @@ library MerklePatriciaProof {
...
-                    if(keccak256(RLP.toData(currentNodeList[1])) == value) {
+                if (pathPtr == path.length) {//leaf node
+                    //check if value of node is same as expected account value
+                    if (keccak256(RLP.toData(currentNodeList[1])) == value) {
                         return true;
                     } else {
                         return false;
                     }
                 }
                 //extension node ... test if means that it is empty value
```
benjaminbollen: for extension node the value must be empty but whether it is an extension node is encoded by the hex-prefix encoding of the path in the node

PR #139, opened 2018-05-23
[MOSA 57] -  Generate Storage Proof on test contract for given mapping key

abhayks1:
[MOSA 57] -  Generate Storage Proof on test contract for given mapping key


abhayks1:
- Generates storage proof using local chain
- Generates storage proof for a contract mapping key

deepesh-kn:
@abhayks1, There are lot of hardcoded value in the code, can we pass the variables as parameters ? DB path can be read from env variable.

PR #143, opened 2018-06-01
Sarvesh/mosa 60 adjust variable position

sarvesh-ost:
1. Refactor contracts to follow the valid sequence. 

2. Verify "is" contract is not breaking the sequence. For example, OpenSTUtility is Hasher, OpsManaged, STPrimeConfig

3. Write a test that will check that condition is not violated- 
               Used solidity parser used in solium linter to parse sol file and validated sequence of variable declaring is correct.


contracts/OpenSTUtility.sol
```diff
@@ -574,31 +576,6 @@ contract OpenSTUtility is Hasher, OpsManaged, STPrimeConfig {
         return uuids.length;
     }
 
-    /*
-     *  Registrar functions
-     */
-    /// @dev for v0.9.1 tracking Ethereum mainnet on the utility chain
-    ///      is not a required feature yet, so the core is simplified
-    ///      to uint256 valueChainId as storage on construction
-    // function addCore(
```
benjaminbollen: why is `addCore` removed?

benjaminbollen: it was dead-code until v0.9.3; so now either this function should not be removed or the constructor needs to have a setter for the core contract that represents OpenSTValue

test/ContractStructureTest.js
```diff
@@ -0,0 +1,94 @@
+let SolidityParser = require("solparse");
+
+
+//todo We can extend this test for all the solidity contracts. But it's out of scope for now
+const contracts = [{
+  name: "OpenSTUtility.sol",
```
benjaminbollen: please correct indentation

benjaminbollen:
Core is added in MOSA-63

PR #147, opened 2018-06-04
Sarvesh/mosa 86 revise return

sarvesh-ost:
Returning relevant data from Gate methods.

### DIFF
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -95,7 +95,7 @@ contract Gate is ProtocolVersioned, Owned {
         uint256 _amount,
         address _beneficiary)
         external
-        returns (bool isSuccess)
+        returns (bool /* success */)
     {
 
         require(_amount > uint256(0));
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -95,7 +95,7 @@ contract Gate is ProtocolVersioned, Owned {
         uint256 _amount,
         address _beneficiary)
         external
-        returns ( bool /* success */)
+        returns ( uint256 /* amount */)
     {
 
         require(_amount > uint256(0));
@@ -115,14 +115,14 @@ contract Gate is ProtocolVersioned, Owned {
 
         emit StakeRequested(msg.sender, _amount, _beneficiary);
 
-        return true;
+        return _amount;
     }
 
 
     /// @dev In order to revert stake request the msg.sender should be the staker
     function revertStakeRequest()
         external
-        returns (bool /* success */)
+        returns (uint256 /* amount */)
     {
         // only staker can do revertStakeRequest, msg.sender == staker
         StakeRequest storage stakeRequest = stakeRequests[msg.sender];
@@ -140,12 +140,12 @@ contract Gate is ProtocolVersioned, Owned {
 
         emit StakeRequestReverted(msg.sender, stakeRequestAmount);
 
-        return true;
+        return stakeRequestAmount;
     }
 
     function rejectStakeRequest(address _staker, uint8 _reason)
         external
-        returns (bool /* success */)
+        returns (uint256 /* amount */)
     {
         // check if the caller is whitelisted worker
         require(workers.isWorker(msg.sender));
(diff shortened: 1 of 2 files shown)


contracts/Gate.sol
```diff
@@ -115,14 +115,14 @@ contract Gate is ProtocolVersioned, Owned {
...
-        return true;
+        return _amount;
     }
 
 
     /// @dev In order to revert stake request the msg.sender should be the staker
     function revertStakeRequest()
         external
-        returns (bool /* success */)
+        returns (uint256 /* amount */)
```
deepesh-kn: We can now declare ``` uint256 stakeRequestAmount ``` in the return itself,  and remove it from line number 138

deepesh-kn: Similarly for ``` rejectStakeRequest ```, ``` processStaking ``` , ``` revertStaking ```

contracts/Gate.sol
```diff
@@ -224,7 +224,7 @@ contract Gate is ProtocolVersioned, Owned {
     bytes32 _stakingIntentHash,
     bytes32 _unlockSecret)
     external
-    returns (bool /* success */)
+    returns (uint /* amount */)
```
benjaminbollen: same comment as @deepesh-kn here, remove L265 and declare L227 as
`returns (uint256 stakeRequestAmount)`
note to prefer `uint256` over `uint`,
note that the suggested name `stakeRequestAmount`

contracts/Gate.sol
```diff
@@ -238,6 +238,7 @@ contract Gate is ProtocolVersioned, Owned {
 
     StakeRequest storage stakeRequest = stakeRequests[staker];
 
+
```
benjaminbollen: remove white line

contracts/Gate.sol
```diff
@@ -259,18 +260,18 @@ contract Gate is ProtocolVersioned, Owned {
       //Transfer bounty amount to the msg.sender account
       require(OpenSTValueInterface(openSTProtocol).valueToken().transfer(msg.sender, bounty));
     }
-
+      uint256 stakeProcessedAmount = stakeRequest.amount;
     // delete the stake request from the mapping storage
     delete stakeRequests[staker];
 
-    return true;
+    return stakeProcessedAmount;
```
benjaminbollen: `return stakeRequestAmount`

contracts/Gate.sol
```diff
@@ -259,18 +260,18 @@ contract Gate is ProtocolVersioned, Owned {
...
   }
 
 
   function revertStaking(
     bytes32 _stakingIntentHash)
     external
-    returns (bool /* success */)
+    returns (uint256 /* amount */)
```
benjaminbollen: similar comment; name of variable should remain `stakeRequestAmount` to be consistent with other two functions

test/Gate.js
```diff
@@ -58,7 +58,9 @@ contract('Gate', function(accounts) {
     if (isSuccessCase) {
       // success case
       let requestStakeResult = await gate.requestStake.call(amount, beneficiary, {from: staker});
-      assert.equal(requestStakeResult, isSuccessCase);
+
```
benjaminbollen: please remove white lines

contracts/Gate.sol
```diff
@@ -115,14 +116,13 @@ contract Gate is ProtocolVersioned, Owned {
 
         emit StakeRequested(msg.sender, _amount, _beneficiary);
 
-        return _amount;
```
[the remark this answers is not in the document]
sarvesh-ost: @benjaminbollen 
Your above comment "same request as @deepesh-kn had", I thought you want to remove explicit returns. 
Just to be clear, we will use returns(uint256 stakeRequestAmount) but will also return explicitly?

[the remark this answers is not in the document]
benjaminbollen: it is redundant, but I feel spooked out by a function without a visible return at the end (what ever form that takes) :), so yes please both define it in the function definition (and doc comments later), and at the end of the function

contracts/Gate.sol
```diff
@@ -95,7 +95,7 @@ contract Gate is ProtocolVersioned, Owned {
         uint256 _amount,
         address _beneficiary)
         external
-        returns ( bool /* success */)
+        returns (bool isSuccess)
```
benjaminbollen: please don't declare `isSuccess` as a variable if unused;
prefer `return (bool /* success */)`

sarvesh-ost: OK, It was added to make all the method signatures look consistent, also it will tell what is meaning of bool return type(here it tells if the operation is successful)

PR #150, opened 2018-06-04
Intents Mapping 

noslav:
* Declares a new mapping `intents` in OpenSTValues.sol
* Adds hashIntenKey fn to Hasher.sol for key in intents mapping 
* Adds to intents mapping in Stake fn of OpenSTValue.sol
* Removes from intents mapping in processStaking, revertStaking fn of OpenSTValue.sol
* Updates `StakingIntentDeclared` event for `intentKeyHash` and relevant tests in `OpenSTValue.js`


sarvesh-ost:
I was looking into this PR out of curiosity.
I think we need to remove the entry from intents mapping during revertStaking as well. 

@benjaminbollen  what is your view on this?

contracts/HasIntents.sol
```diff
@@ -0,0 +1,32 @@
...
+//
+// ----------------------------------------------------------------------------
+// Common: HasIntents
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+
+contract HasIntents {
```
sarvesh-ost: Is this contract name intentional or it's HashIntents?

noslav:
@sarvesh-ost IIUC, the process of unstaking and redeeming would be separate tickets, in which entries would be removed from `intents` mapping.

contracts/HasIntents.sol
deepesh-kn: I have the same question. BTW why not just Intents ?

contracts/HasIntents.sol
```diff
@@ -0,0 +1,32 @@
...
+    /*
+     * Mappings
+     */
+    mapping(bytes32 /* intentHash */ => bytes32) internal intents;
```
deepesh-kn: Just my views: 

I believe that this will be accessed from outside, or interaction layer can call this from outside to check if the value is stored against the key, before generating the proof data.
Hmm...not sure as of now if this should be public or internal. May be this can be changed if needed as we work upon later.

contracts/Hasher.sol
```diff
@@ -93,4 +93,16 @@ contract Hasher {
 			_unlockHeight,
 			_hashLock);
 	}
+
+	function hashIntentKey(
+		address _staker,
```
deepesh-kn: More generic name for Staker ?
Planning to use this contract for OpenSTUtility too to store the redemptionIntentHash.

sarvesh-ost:
I think revertStaking happens in staking flow itself not in redemption. If Stake request is not yet processed, revertStaking can be called. 

This is fine if we are planning to address this in a separate ticket.

contracts/HasIntents.sol
noslav: I kept it internal for the reason that I only expected it to be used by the OpenSTValue contract, but if this will be called on directly from the interaction layer as you have pointed out, then it makes sense to change its type to Public. Will do.

contracts/Hasher.sol
noslav: Are you requesting me to change this to a more generic term and replace `_staker`? Would you have any suggestions? How about simply using `_account` ?

deepesh-kn: Yes !!! sorry for not being clear while commenting.  ``` _account ``` sounds good to me.

contracts/HasIntents.sol
noslav: Initially, I had given it exactly that name, but on the argument that I made above regarding its purpose based on @benjaminbollen advice I changed it to `HasIntents`.

noslav:
@sarvesh-ost  could you kindly point out to me on which line(s) in the contract, `revertStaking` is happening in the `stake` fn?

contracts/HasIntents.sol
benjaminbollen: I take the blame for that, but I neither like `MyContract is Intents`, or `MyContract is HasIntents`; we're misusing the inheritance model here a bit.  I suggested `HasIntents` in line with `HasWorkers` where we also produce bad english to express that it has an object, rather than is a thing...
improvement suggestions for both are very much welcome

benjaminbollen:
@sarvesh-ost you are correct that the intents mapping must also be cleared out on `revertStaking`

@noslav `revertStaking` is a different function https://github.com/OpenSTFoundation/openst-protocol/blob/master/contracts/OpenSTValue.sol#L266

contracts/HasIntents.sol
noslav: @benjaminbollen could we use `MapsIntents` or `MapIntents`?

noslav:
@benjaminbollen  I was indeed aware of that, but according to the meeting notes that I have from that discussion, it was decided that those would be separate tickets? I decided not to increase the scope of the tickets by myself.

noslav:
@sarvesh-ost apologies on misinterpreting your comment earlier, the comment was very helpful!!  👍

contracts/OpenSTValue.sol
```diff
@@ -222,6 +223,8 @@ contract OpenSTValue is OpsManaged, Hasher {
             unlockHeight: unlockHeight,
             hashLock:     _hashLock
         });
+        /** Add intent hash to intents mapping */
```
benjaminbollen: correct comment type, and insert a whiteline above

this is an important line in the code, so it deserve a more elaborate comment, eg:
"Store the staking intent hash directly in storage of OpenSTValue so that a Merkle proof can be generated for active staking intents"

contracts/OpenSTValue.sol
```diff
@@ -256,9 +259,12 @@ contract OpenSTValue is OpsManaged, Hasher {
 
         emit ProcessedStake(stake.uuid, _stakingIntentHash, stakeAddress, stake.staker,
             stake.amountST, stake.amountUT, _unlockSecret);
+        /** Remove intent hash from intents mapping */
+        delete intents[hashIntentKey(stake.staker, stake.nonce)];        
 
         delete stakes[_stakingIntentHash];
 
+
```
benjaminbollen: don't introduce this whiteline

contracts/OpenSTValue.sol
```diff
@@ -256,9 +259,12 @@ contract OpenSTValue is OpsManaged, Hasher {
 
         emit ProcessedStake(stake.uuid, _stakingIntentHash, stakeAddress, stake.staker,
             stake.amountST, stake.amountUT, _unlockSecret);
+        /** Remove intent hash from intents mapping */
```
benjaminbollen: correct comment type, and insert whiteline above; and to be inline with your style in `revertStaking` remove the whiteline between the deletes

contracts/OpenSTValue.sol
```diff
@@ -73,8 +72,9 @@ contract OpenSTValue is OpsManaged, Hasher {
     uint256 private constant BLOCKS_TO_WAIT_LONG = 80667;
     // ~1hour, assuming ~15s per block
     uint256 private constant BLOCKS_TO_WAIT_SHORT = 240;
-
-
+    
+    /** capture intents */
```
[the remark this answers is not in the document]
noslav: should I be using `//` or `///`?

contracts/OpenSTValue.sol
noslav: Done!

contracts/OpenSTValue.sol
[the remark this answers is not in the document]
noslav: fixed with `//`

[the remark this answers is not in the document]
noslav: Corrected to `//`

contracts/OpenSTValue.sol
```diff
@@ -223,6 +224,10 @@ contract OpenSTValue is OpsManaged, Hasher {
             hashLock:     _hashLock
         });
 
+        // store the staking intent hash directly in storage of OpenSTValue 
+        // so that a Merkle proof can be generated for active staking intents
+        intents[hashIntentKey(_staker, nonce)] = stakingIntentHash;
```
abhayks1: hashIntentKey(_staker, nonce) it can be taken in a variable and should be part of event. Platform will need this to generate storage proof.

abhayks1: It should be part of StakingIntentDeclared event.

contracts/OpenSTValue.sol
```diff
@@ -73,9 +72,10 @@ contract OpenSTValue is OpsManaged, Hasher {
     uint256 private constant BLOCKS_TO_WAIT_LONG = 80667;
     // ~1hour, assuming ~15s per block
     uint256 private constant BLOCKS_TO_WAIT_SHORT = 240;
-
-
-    /// register the active stakes and unstakes
+    
+    // storage for staking intent hash of active staking intents
+    mapping(bytes32 /* intentHash */ => bytes32) public intents;
```
abhayks1: We need to define index position of intents mapping in OpenSTUtility contract.
There needs to be unit test cases for the index position.

Index position of intents mapping is needed to generate storage proof.

contracts/OpenSTValue.sol
noslav: Done!

contracts/OpenSTValue.sol
noslav: Yes, this was recommended by @benjaminbollen to be made into a separate ticket IIRC. I can pick up it up.

abhayks1:
As discussed with @noslav we should create ticket for below change:

We need to define index position of intents mapping in OpenSTUtility contract.
There needs to be unit test cases for the index position.
Index position of intents mapping is needed to generate storage proof.

cc: @pgev

abhayks1:
@noslav please check the test cases and update accordingly

noslav:
@abhayks1 incorporated all requested changes (except for index check on `intents` mapping), added the relevant updates to the test cases emitting the `StakingIntentDeclared` event.

PR #152, opened 2018-06-05
Sarvesh/mosa 85 make stake request iterable

sarvesh-ost:
1. Created Library contract for address array 
2. Adding staker address in requestStake method
3. Removing staker address in revertStakeRequest, rejectStakeRequest, processStaking and revertStaking
4. Create a function to return length of array 
5. Test cases to support above changes

### DIFF
--- a/contracts/AddressArrayLib.sol
+++ b/contracts/AddressArrayLib.sol
@@ -1,0 +1,40 @@
+pragma solidity ^0.4.23;
+
+library AddressArrayLib {
+
+    struct AddressArray {
+        address[] addresses;
+    }
+
+    function add(AddressArray storage array, address value) internal returns (uint){
+        array.addresses.push(value);
+        return array.addresses.length;
+    }
+
+    function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
+        uint i = find(array, value);
+        removeByIndex(array, i);
+        return array.addresses.length;
+    }
+
+
+    function find(AddressArray storage array, address value) internal view returns (uint) {
+        uint i = 0;
+        while (array.addresses[i] != value) {
+            i++;
+        }
+        return i;
+    }
+
+    function removeByIndex(AddressArray storage array, uint i) internal view {
+        while (i < array.addresses.length - 1) {
+            array.addresses[i] = array.addresses[i + 1];
+            i++;
+        }
+        array.addresses.length--;
+    }
+
+    function length(AddressArray storage array) internal view returns(uint){
+        return array.addresses.length;
+    }
+}
--- a/contracts/AddressArrayLib.sol
+++ b/contracts/AddressArrayLib.sol
@@ -1,23 +1,45 @@
 pragma solidity ^0.4.23;
 
+// Copyright 2018 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+// Common : Address Array
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+
 library AddressArrayLib {
 
     struct AddressArray {
         address[] addresses;
     }
 
-    function add(AddressArray storage array, address value) internal returns (uint){
+    //@dev Add element to array
+    function add(AddressArray storage array, address value) internal view returns (uint){
         array.addresses.push(value);
         return array.addresses.length;
     }
-
+    //@dev Remove element from array by value
     function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
         uint i = find(array, value);
         removeByIndex(array, i);
         return array.addresses.length;
     }
-
-
+    //@dev search element in array and return index
     function find(AddressArray storage array, address value) internal view returns (uint) {
         uint i = 0;
         while (array.addresses[i] != value) {
@@ -25,7 +47,7 @@ library AddressArrayLib {
         }
         return i;
     }
-
+    //@dev remove element for array using index
     function removeByIndex(AddressArray storage array, uint i) internal view {
         while (i < array.addresses.length - 1) {
             array.addresses[i] = array.addresses[i + 1];
@@ -33,7 +55,7 @@ library AddressArrayLib {
         }
         array.addresses.length--;
     }
-
+    //@dev return length of array
     function length(AddressArray storage array) internal view returns(uint){
         return array.addresses.length;
     }
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -27,9 +27,10 @@ import "./OpenSTValueInterface.sol";
 import "./EIP20Interface.sol";
 import "./Owned.sol";
 import "./WorkersInterface.sol";
+import "./AddressArrayLib.sol";
 
 contract Gate is ProtocolVersioned, Owned {
-
+    using AddressArrayLib for AddressArrayLib.AddressArray;
     /*
      * Events
      */
@@ -69,7 +70,7 @@ contract Gate is ProtocolVersioned, Owned {
     // utility token UUID
     bytes32 public uuid;
 
-
+    AddressArrayLib.AddressArray stakers;
     /*
      *  Public functions
      */
@@ -113,6 +114,8 @@ contract Gate is ProtocolVersioned, Owned {
             unlockHeight: 0
         });
 
+        stakers.add(msg.sender);
+
         emit StakeRequested(msg.sender, _amount, _beneficiary);
 
         return true;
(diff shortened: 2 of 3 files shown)


contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    //@dev Add element to array
+    function add(AddressArray storage array, address value) internal view returns (uint){
+        array.addresses.push(value);
+        return array.addresses.length;
+    }
+    //@dev Remove element from array by value
+    function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
+        uint i = find(array, value);
+        removeByIndex(array, i);
```
deepesh-kn: If the element is not found, then it still calls ``` removeByIndex(array, i); ```  and inside that method it does ``` array.addresses.length--; ``` even if the element was not found.

Need to add check for the i before calling ``` removeByIndex(array, i); ``` ?

deepesh-kn: Or how exactly ``` array.addresses.length--;  ``` works ?

sarvesh-ost: In solidity, Dynamic arrays can be resized in storage (not in memory) by changing the .length property.
If element not found in the array, EVM will throw an exception and revert transaction.  However, I feel we should gracefully revert transaction explicitly.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+// ----------------------------------------------------------------------------
+
+
+library AddressArrayLib {
+
+    struct AddressArray {
+        address[] addresses;
+    }
+
+    //@dev Add element to array
```
pgev: Let's use /** */ styling.
In addition // (/* */) is just an ordinary comment ... /// (/** */) is a document comment.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+    function add(AddressArray storage array, address value) internal view returns (uint){
+        array.addresses.push(value);
+        return array.addresses.length;
+    }
+    //@dev Remove element from array by value
+    function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
+        uint index;
```
pgev: As a coding style I would propose to always initialize variables. We can discuss it.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+
+
+library AddressArrayLib {
+
+    struct AddressArray {
+        address[] addresses;
+    }
+
+    //@dev Add element to array
+    function add(AddressArray storage array, address value) internal view returns (uint){
```
pgev: Again something I would like to bring into discussion.
Should it be internal or private? Do we assume that we are going to inherit and override the function?

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+        array.addresses.push(value);
+        return array.addresses.length;
+    }
+    //@dev Remove element from array by value
+    function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
+        uint index;
+        bool isElementPresent;
+        (isElementPresent, index) = find(array, value);
+        require(isElementPresent, "Element doesn't exist in Array");
```
pgev: Do we really want the function to fail and throw an exception if element does not exist? There is no catching in solidity :)

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+        removeByIndex(array, index);
+        return array.addresses.length;
+    }
+    //@dev search element in array and return index
+    function find(AddressArray storage array, address value) internal view returns (bool, uint) {
+        uint i = 0;
+        bool isElementPresent = false;
+        while (!(isElementPresent = (array.addresses[i] == value)) && i < array.addresses.length) {
```
pgev: Will this through an exception if the element is not in array?

contracts/AddressArrayLib.sol
sarvesh-ost: Private methods are not accessible outside contracts. Methods in this library are consumed outside.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+            i++;
+        }
+        return (isElementPresent, i);
+    }
+    //@dev remove element for array using index
+    function removeByIndex(AddressArray storage array, uint i) internal view {
+        while (i < array.addresses.length - 1) {
```
pgev: This check is tricky in case of empty array we have a -1 on a right side.
Also, we decrease array length always ...

contracts/AddressArrayLib.sol
sarvesh-ost: No, It will not throw the exception, it will return false

contracts/AddressArrayLib.sol
sarvesh-ost: removeByIndex is called in removeByValue, it performs checks  whether values exist or not by find operation. It will be called only if the value exists. We can make this method private just to make sure it's not called for externally.

contracts/AddressArrayLib.sol
sarvesh-ost: I don't see any harm in reverting transaction in this use case if stake entry doesn't exist in the array.
Find method is returning true and false, but to me, it makes sense for removeByValue to revert transaction if the value doesn't exist.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,66 @@
...
+    function find(AddressArray storage array, address value) internal view returns (bool, uint) {
+        uint i = 0;
+        bool isElementPresent = false;
+        while (!(isElementPresent = (array.addresses[i] == value)) && i < array.addresses.length) {
+            i++;
+        }
+        return (isElementPresent, i);
+    }
+    //@dev remove element for array using index
+    function removeByIndex(AddressArray storage array, uint i) private view {
```
pgev: 1. If array is empty this could run forever (till gas is consumed).
2. If index is bigger then array length we still decrease array length.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+    struct AddressArray {
+        address[] addresses;
+    }
+
+    /*  @title -  Add element to array
+     *  @Param - array
+     *  @Param - value which will be added to array
+     *  @Returns: new length of array
+     */
+    function add(AddressArray storage array, address value) internal view returns (uint){
```
pgev: add -> push ?

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+        require(value != address(0), "value should be non zero");
```
pgev: Not sure about this limitation. It's assumed to be a generic library ...

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    /*  @title -  Remove first occurrence of element from array by value
+     *  @Param - array
+     *  @Param - value which will be removed from
+     *  @Returns: new length of array
+     */
+    function removeByValue(AddressArray storage array, address value) internal view returns (uint) {
+        require(value != address(0), "value should be non zero");
+
+        uint256 index;
```
pgev: Not a huge fun of uninitialized variables. As a suggestion:

uint256 index = 0;
bool isElementPresent = false;

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+        bool isElementPresent;
+
+        (isElementPresent, index) = find(array, value);
+
+        require(isElementPresent, "Element doesn't exist in Array");
```
pgev: We should not throw an exception if element does not exist. We could return false indicating that no element was removed.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    /*  @title -  Add element to array
+     *  @Param - array
+     *  @Param - value which will be added to array
+     *  @Returns: new length of array
+     */
+    function add(AddressArray storage array, address value) internal view returns (uint){
+        require(value != address(0), "value should be non zero");
+
+        array.addresses.push(value);
```
pgev: return array.addresses.push(value);

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+
+        (isElementPresent, index) = find(array, value);
+
+        require(isElementPresent, "Element doesn't exist in Array");
+
+        removeByIndex(array, index);
+        return array.addresses.length;
+    }
+
+    /*  @title - Search element in array
```
pgev: Documentation should state that the first occurrence is going to be returned.

contracts/AddressArrayLib.sol
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /*  @title - Remove element in array by index
+     *  @Param - array
+     *  @Param - index of element
+     *  @Returns - new length of array
+     */
+    function removeByIndex(AddressArray storage array, uint256 i) private view {
+        require(i >= 0, "Index should be greater than zero");
+        require(i < array.addresses.length, "Index should be less than length");
```
pgev: Can we specify for example what is the passed index, what is an address length. I mean more contextual info to make the error message more informative?

sarvesh-ost: Appending uint to string is bit complex thing. It requires to remover each digit of uint and convert into bytes and concatenate.  I feel we shouldn't pollute library logic for the sake of require reason.  It doesn't add much value. Let me know if have another opinion.

benjaminbollen:
For the data to be iterable we don't need this machinery.

I don't understand why the library functions have `view` as visibility.  There are other remarks to be made, but we should first align what the requirements are.

sarvesh-ost:
@benjaminbollen  Thanks for the feedback. Let's discuss this in the call.

PR #154, opened 2018-06-05
[MOSA-96] - RLP.sol - Use of the "var" keyword is deprecated

abhayks1:
Solution is to use strictly typed variables

### DIFF
--- a/contracts/proof/RLP.sol
+++ b/contracts/proof/RLP.sol
@@ -189,8 +189,7 @@ library RLP {
  /// @return Array of RLPItems.
  function toList(RLPItem memory self) internal pure returns (RLPItem[] memory list) {
      require(isList(self));
-     uint numItems;
-     numItems = items(self);
+     uint numItems = items(self);
      list = new RLPItem[](numItems);
      Iterator memory it = iterator(self);
      uint idx;
--- a/contracts/proof/RLP.sol
+++ b/contracts/proof/RLP.sol
@@ -33,8 +33,8 @@ library RLP {
 
  function next(Iterator memory self) internal pure returns (RLPItem memory subItem) {
      require(hasNext(self));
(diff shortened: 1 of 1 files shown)


contracts/proof/RLP.sol
```diff
@@ -33,8 +33,8 @@ library RLP {
 
  function next(Iterator memory self) internal pure returns (RLPItem memory subItem) {
      require(hasNext(self));
-     var ptr = self._unsafe_nextPtr;
-     var itemLength = _itemLength(ptr);
+     uint256 ptr = self._unsafe_nextPtr;
```
deepesh-kn: Do line# 29 needs to change to uint256 ?

pgev: uint and int are aliases for uint256 and int256, respectively.

abhayks1: Yes because _itemLength takes ptr as input which is uint256.

contracts/proof/RLP.sol
```diff
@@ -187,9 +189,10 @@ library RLP {
  /// @return Array of RLPItems.
  function toList(RLPItem memory self) internal pure returns (RLPItem[] memory list) {
      require(isList(self));
-     var numItems = items(self);
+     uint numItems;
```
pgev: Why not declare and define at the same time?

abhayks1: you are right.
Made the change.

benjaminbollen:
> For Below syntax:
> (uint a, uint b, uint c) = f();
> It will be taken care with below ticket.
> https://ostdotcom.atlassian.net/browse/MOSA-114

for community / public s clarity: this cleaner syntax is possible in solidity v0.4.24 and our internal ticket MOSA-114 is to upgrade solidity to v0.4.24;

PR #155, opened 2018-06-06
rearranging variable and mapping as per standard. Also adding Gate co…

sarvesh-ost:
I. Rearranging variable positions in the below-mentioned sequence 
   Events => Constant => Mapping  => Variables  => Arrays => Struct

II.Add Gate.sol in contract structure test

### DIFF
--- a/test/ContractStructureTest.js
+++ b/test/ContractStructureTest.js
@@ -10,8 +10,13 @@ const contracts = [
   {
     name: "OpenSTValue.sol",
     path: "./contracts/OpenSTValue.sol"
+  },
+  {
+    name: "Gate.sol",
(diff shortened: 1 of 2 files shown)


test/ContractStructureTest.js
```diff
@@ -10,8 +10,13 @@ const contracts = [
   {
     name: "OpenSTValue.sol",
     path: "./contracts/OpenSTValue.sol"
+  },
+  {
+    name: "Gate.sol",
+    path: "./contracts/Gate.sol"
   }
 ];
+const validSequence = ["event", "constant", "mapping", "variable", "array", "struct"];
```
pgev: Not seeing the reason moving this out from function declaration.

PR #156, opened 2018-06-08
Gulshan/mosa 49

gulshanvasnani:


package.json
```diff
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "truffle": "^4.1.8",
-    "ganache-cli": "^6.0.3",
+    "ganache-cli": "^6.1.0",
     "abi-decoder": "^1.0.9",
     "ethereumjs-testrpc": "^6.0.3"
```
jasonklein: Unless you disagree, I think this should be removed.

test/OpenSTUtility.js
```diff
@@ -363,7 +360,7 @@ contract('OpenSTUtility', function(accounts) {
 			var redeemReturns = await openSTUtility.redeem.call(checkBtUuid, redeemAmountUT, 2, redeemBeneficiary, lockR.l, { from: redeemer });
 
             // call block number is one less than send block number
```
jasonklein: Global comment for where this inline comment appears: I think this inline comment no longer applies

PR #160, opened 2018-06-19
Added comments for Gate.sol. MOSA-99

deepesh-kn:
Github issue ref: #164
Added comments using natspec format for each functions in Gate contract.

_MOSA-99_


deepesh-kn:
undone the code change at  line 293 in Gate.sol for processStaking. 
This function can be called by anyone and we have proper checks in line 318.
```
// check if the caller is whitelisted worker
    require(workers.isWorker(msg.sender));
```
I remember that this change is already done in some other branch. I will check and if its not there we can create an issue for that fix.

contracts/Gate.sol
```diff
@@ -165,8 +216,22 @@ contract Gate is ProtocolVersioned, Owned {
         return stakeRequestAmount;
     }
 
-    /// @dev In order to accept stake the staker needs to approve gate contract for bounty amount.
-    ///      This can be called only by whitelisted worker address
+    /**
+      * @notice external function to accept requested stake.
+      *
+      * @dev In order to accept stake the staker needs to approve gate contract for bounty amount.
```
jasonklein: I think this comment needs some improvement:
- it is not clear to me what is doing the accepting in the clause: "In order to accept stake"
- I think it's the worker/the caller of this function that needs to approve the Gate contract for the bounty, not the staker
- I would switch the last two lines, "OpenSTProtocol is approved..." and "Bounty amount is transferred...", so that they appear in the same order in which they occur in the function
- while I'm not sure if it's better to refer to "OpenSTProtocol" or "OpenSTValue", the variable name from `ProtocolVersioned` is lower-case, "openSTProtocol", so suggest sticking with that in the documentation comment

contracts/Gate.sol
```diff
@@ -214,92 +279,112 @@ contract Gate is ProtocolVersioned, Owned {
         return (amountUT, nonce, unlockHeight, stakingIntentHash);
     }
 
+    /**
+      * @notice external function to process staking.
+      *
+      * @dev Bounty amount is to msg.sender if its not whitelisted worker
```
jasonklein: Please add a word like "transferred" or "sent" between "amount is" and "to" in both lines here.

"its" --> "it's"

contracts/Gate.sol
jasonklein: There's a comment on what is currently L257 that has a "form" that should be "from": "// Transfer bounty amount form worker to gate contract"

contracts/Gate.sol
```diff
@@ -214,92 +279,112 @@ contract Gate is ProtocolVersioned, Owned {
...
-    returns (uint256 stakeRequestAmount)
-  {
+        return stakeRequestAmount;
+      }
+
+    /**
+      * @notice external function to revert staking.
+      *
+      * @dev Caller must be the whitelisted worker.
+      *      Staked amount is transferred the staker address.
```
jasonklein: Please add "to" between "transferred" and "the staker".

contracts/Gate.sol
```diff
@@ -214,92 +279,112 @@ contract Gate is ProtocolVersioned, Owned {
...
+      *      Bounty amount is to workers contract.
```
jasonklein: Please add "transferred" between "amount is" and "to".

contracts/Gate.sol
```diff
@@ -28,14 +28,27 @@ import "./EIP20Interface.sol";
 import "./Owned.sol";
 import "./WorkersInterface.sol";
 
+/**
+ *	@title Gate contract which implements ProtocolVersioned, Owned
+ *
+ *	@notice Gate contract is staking gate that seperates the concerns of staker and staking processor.
+ *      Stake process is executed through Gate contract rather than directly with the protocol contract.
+ *      The Gate contract will server the role of staking account rather than an external account.
```
jasonklein: "server" == "serve"

jasonklein:
LGTM! 🚀

Issue #166, opened 2018-06-27
Replace OpenSTProven and ProofVerificationSkipped events by OpenSTProven with additional parameter


abhayks1:
Currently, there are two events: OpenSTProven and ProofVerificationSkipped that are fired within proveOpenST function. The scope of this ticket to replace both by one event: OpenSTProven.

Current events signature:
event OpenSTProven(uint256 blockHeight, bytes32 storageRoot, bytes32 hashedAccount);
event ProofVerificationSkipped(uint256 blockHeight, bytes32 storageRoot);

Please, replace them by:
event OpenSTProven(uint256 blockHeight, bytes32 storageRoot, bool wasAlreadyProved);

Reference - https://ostdotcom.atlassian.net/browse/MOSA-129

PR #167, opened 2018-06-27
[MOSA-129] Replace OpenSTProven and ProofVerificationSkipped events by OpenSTProven with additional parameter	

abhayks1:
- Removed event ProofVerificationSkipped 
- Removed parameter hashedAccount from event signature of OpenSTProven      
- Added parameter wasAlreadyProved in event signature of OpenSTProven


abhayks1:
Resolves issue #166

Issue #178, opened 2018-07-04
Gate::revertStaking should not require caller to be whitelisted worker


deepesh-kn:
Currently, revertStaking() function of Gate.sol contract requires the caller to be whitelisted worker.This is not needed anymore and should be removed.

Test case should be introduced that check:

- Check that stake exists and is unlocked, otherwise fail.
- Check that staked amount is transferred back to staker.
- If it's a whitelisted worker, then bounty is sent from OpenSTValue contract to Workers contract.
- If it's non whitelisted worker, then bounty is sent from OpenSTValue to msg.sender.
- Stake request should be deleted afterwards from Gate contract.

_MOSA-127_

deepesh-kn:
**Update:** The bounty will be always transferred to workers contract.

jasonklein:
Resolved by #180 .

PR #180, opened 2018-07-04
Gate::revertStaking should not require caller to be whitelisted worker [MOSA-127]

deepesh-kn:
`Gate.revertStaking` now can be called by any account, not just whitelisted worker.

GH issue: #178 
Branch: deepesh/MOSA-127

Please note:
It is mentioned in the gh issue #178 : _"If it's non whitelisted worker, then bounty is sent from OpenSTValue to msg.sender."_
This is not implemented in this PR.
user can stake amount, wait for unlock time and then revert it. User get the bounty amount :)
Please let me know if my understanding is correct.

### DIFF
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -334,8 +334,7 @@ contract Gate is ProtocolVersioned, Owned {
     /**
       * @notice external function to revert staking
       *
-      * @dev Caller must be the whitelisted worker
-      *      Staked amount is transferred to the staker address
+      * @dev Staked amount is transferred to the staker address
       *      Bounty amount is transferred to workers contract
       *
       * @param _stakingIntentHash staking intent hash
@@ -347,9 +346,6 @@ contract Gate is ProtocolVersioned, Owned {
         external
         returns (uint256 stakeRequestAmount)
       {
-
-        // check if the caller is whitelisted worker
-        require(workers.isWorker(msg.sender));
 
         require(_stakingIntentHash != bytes32(0));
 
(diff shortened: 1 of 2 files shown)


contracts/Gate.sol
```diff
@@ -348,9 +347,6 @@ contract Gate is ProtocolVersioned, Owned {
         returns (uint256 stakeRequestAmount)
```
jasonklein: This should be amountST

contracts/Gate.sol
```diff
@@ -366,20 +362,20 @@ contract Gate is ProtocolVersioned, Owned {
         // check if the stake request was accepted
         require(stakeRequest.hashLock != bytes32(0));
 
-        bytes32 uuidR = bytes32(0);
         uint256 amountST = uint256(0);
         address stakerAddress = address(0);
-        (uuidR, amountST, stakerAddress) = OpenSTValueInterface(openSTProtocol).revertStaking(_stakingIntentHash);
+        (, amountST, stakerAddress) = OpenSTValueInterface(openSTProtocol).revertStaking(_stakingIntentHash);
```
jasonklein: I think you can define `address stakerAddress` here, so I think you can delete the line above.

contracts/Gate.sol
```diff
@@ -366,20 +362,20 @@ contract Gate is ProtocolVersioned, Owned {
...
 
-        // check if the stake address is not 0
-        require(stakerAddress != address(0));
+        // check if the stake address is correct
+        require(stakerAddress == staker);
```
jasonklein: It should be the case that this has to be true, so does this compile with an `assert`?

contracts/Gate.sol
```diff
@@ -366,20 +362,20 @@ contract Gate is ProtocolVersioned, Owned {
...
+
+        require(amountST == stakeRequest.amount);
```
jasonklein: Like above, it should be the case that this has to be true, so does this compile with an `assert`?

contracts/Gate.sol
deepesh-kn: done !

contracts/Gate.sol
deepesh-kn: Moving to inline, gives compilation error.

contracts/Gate.sol
deepesh-kn: Done !

contracts/Gate.sol
deepesh-kn: Done !

Issue #184, opened 2018-07-09
Add test cases for openst-protocol RLP library contract


gulshanvasnani:

jasonklein:
Resolved by #185.

PR #185, opened 2018-07-09
 RLP library test cases[Gulshan/mosa 123]

gulshanvasnani:
Includes test cases for the following functions:
`RLP.toRLPItem`
`RLP.toList`
`RLP.toBytes`
`RLP.toData`

GH issue: #184 
ref: _MOSA-123_


test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+
+        })
+
+        it('should fail when RLP is incorrect with option strict as true', async () => {
+
+            await Utils.expectThrow(rlpMock.toRLPItemStrict.call(dataArray, true));
+
+        })
+
+        it('should pass when RLP list is correct', async () => {
```
jasonklein: Please test the fail case(s), when the provided data is not a list. Can `ret` be `false` in `ret := lt(byte(0, mload(memPtr)), 0xC0)` with a length > 0?

Please also revise the description of this test similarly to description revisions requested above.

contracts/RLPTest.sol
```diff
@@ -0,0 +1,107 @@
...
+     */
+    function toList(
+        bytes rlpEncodedData)
+        pure
+        public
+        returns(uint)
+    {
+        RLP.RLPItem memory item = RLP.toRLPItem(rlpEncodedData);
+        RLP.RLPItem[] memory list =  RLP.toList(item);
+        return list.length;
```
jasonklein: I think this should return the list of sub-items. Returning the length of the list of sub-items does not provide me confidence that `toList` correctly returns an array of the components of `item`.

test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+
+            let dataInNibbles = ethUtil.rlp.encode(dataArray).toString('hex')
+                , lengthOfList = await rlpMock.toList.call('0x' + dataInNibbles);
+            assert.equal(dataArray.length, lengthOfList);
+
+        })
+
+        it('should pass when toBytes is passed with correct rlp data ', async () => {
```
jasonklein: Please revise the description of this test similarly to description revisions requested above (referring to the underlying function that the function in `RLPTest` is testing is different from the previous tests).

test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+
+            let dataInNibbles = ethUtil.rlp.encode(dataArray).toString('hex'),
+                result = await rlpMock.toBytes.call('0x' + dataInNibbles);
+            assert.equal(result.replace("0x", ""), dataInNibbles);
+
+        })
+
+
+        it('should pass when toData is passed with correct rlp data', async () => {
```
jasonklein: Please revise the description of this test similarly to description revisions requested above (referring to the underlying function that the function in `RLPTest` is testing is different from the previous tests).

test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+
+            let data = "1234",
+                dataInNibbles = ethUtil.rlp.encode(data).toString('hex'),
+                result = await rlpMock.toData.call('0x' + dataInNibbles);
+            assert.equal(result.replace("0x", ""), Buffer.from(data).toString('hex'))
+
+        })
+
+        it('should fail when toData is passed with array of rlp data', async () => {
```
jasonklein: Please revise the description of this test similarly to description revisions requested above (referring to the underlying function that the function in `RLPTest` is testing is different from the previous tests). Is an array of RLP data not data? Also, it looks like a hexstring is passed, and not an array.

Please clarify which of the fail cases this is testing and please also test the other fail case (similar question as with `toList`: can `ret` be `false` in `ret := lt(byte(0, mload(memPtr)), 0xC0)` with a length > 0?

test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+            let dataInNibbles = ethUtil.rlp.encode(dataArray).toString('hex'),
+                result = await rlpMock.toBytes.call('0x' + dataInNibbles);
+            assert.equal(result.replace("0x", ""), dataInNibbles);
+
+        })
+
+
+        it('should pass when toData is passed with correct rlp data', async () => {
+
+            let data = "1234",
```
jasonklein: Global comment: please apply a consistent style throughout this file for commas and lists of variable definitions.

test/RLP.js
```diff
@@ -0,0 +1,74 @@
...
+    before(async () => {
+        rlpMock = await RLP.new();
+
+    });
+    let dataArray = ['2', '5', '6'];
+    describe('Test Cases ', async () => {
+
+        it('should pass when RLP data is correct', async () => {
+
+            dataInNibbles = ethUtil.rlp.encode(dataArray).toString('hex')
```
jasonklein: Should this be a constant given that it's used so frequently?

jasonklein:
LGTM

Issue #186, opened 2018-07-10
Implement function setWorkers in Gateway contract


deepesh-kn:
Implement `setWorkers` function in Gateway contract.
This can be only called by owner

ref: MOSA-30

jasonklein:
Resolved by #187 .

PR #187, opened 2018-07-10
Add setWorkers function in gateway contract. MOSA-30

deepesh-kn:
Implements `setWorkers` function in Gateway contract, can be only called by owner.
Includes the setWorker test case.

ref: MOSA-30
fixes: #186

### DIFF
--- a/contracts/Gateway.sol
+++ b/contracts/Gateway.sol
@@ -58,7 +58,7 @@ contract Gateway is ProtocolVersioned, Owned {
       bytes32 _stakingIntentHash);
 
     /** Below event is emitted after successful execution of setWorkers */
-    event WorkersSet(bytes32 _uuid, WorkersInterface _workers);
+    event WorkersSet(WorkersInterface _workers);
 
     /* Storage */
 
@@ -399,7 +399,7 @@ contract Gateway is ProtocolVersioned, Owned {
         workers = _workers;
 
         //Event for workers set
-        emit WorkersSet(uuid, _workers);
+        emit WorkersSet(_workers);
 
         return true;
     }
--- a/contracts/Gateway.sol
+++ b/contracts/Gateway.sol
@@ -386,20 +386,20 @@ contract Gateway is ProtocolVersioned, Owned {
       *
       * @dev only owner can call this
       *
-      * @param _worker workers contract address
+      * @param _workers workers contract address
       *
       * @return success, boolean that specifies status of the execution
       */
-    function setWorkers(WorkersInterface _worker)
+    function setWorkers(WorkersInterface _workers)
         external
         onlyOwner()
         returns (bool /* success */)
     {
-        require(_worker != address(0));
-        workers = _worker;
+        require(_workers != address(0));
+        workers = _workers;
 
         //Event for workers set
-        emit WorkersSet(uuid, _worker);
+        emit WorkersSet(uuid, _workers);
 
         return true;
     }
--- a/contracts/Gateway.sol
+++ b/contracts/Gateway.sol
@@ -57,6 +57,9 @@ contract Gateway is ProtocolVersioned, Owned {
       uint256 _unlockHeight,
       bytes32 _stakingIntentHash);
 
+    /** Below event is emitted after successful execution of setWorkers */
+    event WorkersSet(bytes32 _uuid, WorkersInterface _workers);
+
     /* Storage */
 
     /** Storing stake requests */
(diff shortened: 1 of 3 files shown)


test/Gateway_utils.js
```diff
@@ -161,11 +163,16 @@ module.exports.checkProcessedStakeEvent = (event, _staker, _amount) => {
...
     _amount = new BigNumber(_amount);
   }
-  console.log("event: ",event);
   assert.equal(event.event, "ProcessedStake");
   assert.equal(event.args._staker, _staker);
   assert.equal(event.args._amountST.toNumber(10), _amount.toNumber(10));
 };
 
 
+module.exports.checkWorkersSetEvent = (event, _workerAddress, _uuid) => {
```
jasonklein: `_workerAddress` --> `_workersAddress`

contracts/Gateway.sol
```diff
@@ -57,6 +57,9 @@ contract Gateway is ProtocolVersioned, Owned {
       uint256 _unlockHeight,
       bytes32 _stakingIntentHash);
 
+    /** Below event is emitted after successful execution of setWorkers */
+    event WorkersSet(bytes32 _uuid, WorkersInterface _workers);
```
jasonklein: I don't believe `_uuid` is needed

test/Gateway.js
```diff
@@ -316,6 +319,24 @@ contract('Gateway', function(accounts) {
...
+  const setWorkers = async function (workersAddress, messageSender, isSuccessCase) {
+
+    if (isSuccessCase){
+      let response = await gateway.setWorkers.call(workersAddress, {from: messageSender});
+      assert.equal(response, isSuccessCase);
+
+      let result = await gateway.setWorkers(workersAddress, {from: messageSender});
+      await Gateway_utils.checkWorkersSetEvent(result.logs[0], workersAddress, uuid);
+
+      let workersFromContract = await  gateway.workers.call();
```
jasonklein: Please remove extra space after `await`

test/Gateway.js
```diff
@@ -695,7 +716,23 @@ contract('Gateway', function(accounts) {
...
+    });
+
+    describe('setWorkers', async () => {
+
+        before (async () => {
+            await deployGateway();
+        });
+
+        it('fails to set workers when caller is not owner', async () => {
+            await setWorkers(accounts[8], accounts[2], false)
```
jasonklein: Please add a semi-colon

test/Gateway.js
```diff
@@ -695,7 +716,23 @@ contract('Gateway', function(accounts) {
...
+        });
+
+        it('sucessfully sets workers when caller is owner', async () => {
+            await setWorkers(accounts[8], ownerAddress, true)
```
jasonklein: Please add a semi-colon

deepesh-kn: Done !

test/Gateway.js
deepesh-kn: Done !

test/Gateway.js
deepesh-kn: Done !

contracts/Gateway.sol
deepesh-kn: I have removed it.

I am not aware of the use-case where this will be consumed. I believe that when block scanner will read the receipt, it will get the information about the Gateway address and workers address.  UUID was an additional info that was available to extract from the receipt rather than going to external DB.

So it make sense to not have it for now until we get a use-case for this

test/Gateway_utils.js
deepesh-kn: Done !

contracts/Gateway.sol
```diff
@@ -377,4 +380,27 @@ contract Gateway is ProtocolVersioned, Owned {
...
+      * @param _workers workers contract address
+      *
+      * @return success, boolean that specifies status of the execution
+      */
+    function setWorkers(WorkersInterface _workers)
+        external
+        onlyOwner()
+        returns (bool /* success */)
+    {
+        require(_workers != address(0));
```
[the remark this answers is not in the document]
deepesh-kn: I see your point.
We should discuss this and above comment in group.
Lets discuss this tomorrow.

- I think we should have a proper check in setWorkers to validate that the new address is workers contract address
- To turn off certain functionality, for now setting it to 0 is equivalent to Workers.remove().  We should think of this use-case properly, I am not sure why owner will want to turn-off certain functionality.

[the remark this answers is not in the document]
jasonklein: Let's put a pin in the idea of a remove function (note it for later), which I would argue is an enhancement over permitting setting to 0; whereas my concern with setting to 0 is why have the code if we don't have a reason for having the code, and that removing gives the additional benefit of increasing flexibility for the owner.

As far as why an owner will want to turn-off certain functionality, that enters the realm of "business decisions" (on which I prefer not to theorize or opine) and so I will always side with giving more flexibility unless there's a protocol-concern against it.

PR #213, opened 2018-07-25
Mosa 156/return remainder token to staker

sarvesh-ost:
1. Created Library for token conversion calculations
2. In gate consuming only ST which can be converted to BT.

#219


contracts/Gateway.sol
```diff
@@ -101,14 +107,16 @@ contract Gateway is ProtocolVersioned, Owned {
...
+        conversionRate = _conversionRate;
+        conversionRateDecimals = _conversionRateDecimals;
     }
 
     /**
      *  @notice External function requestStake.
      *
      *  @dev In order to request stake the staker needs to approve Gateway contract for stake amount.
      *       Staked amount is transferred from staker address to Gateway contract.
+     *   Only part of stake amount is consumed which can minted to BT without any loss
```
jasonklein: I think this statement is misleading, because (a) it's possible that all of the amount will be consumed—a clunkier, but accurate, description would be the _maximum stakeable amount given the conversion rate_ (and specifically not `conversionRate`) is consumed; and (b) this is not loss-less, but the amount of loss is at most 1 OST Wei, which we think is acceptable. Additionally, I think it might be confusing that the documentation refers to "stake[d] amount" now that the event includes an element, `_stakedAmount`. Suggest referring to "requested stake amount" or "requested amount" instead; please update the entirety of the `@dev` for this function accordingly.

Please also align this last sentence with the rest of the `@dev` comment.

contracts/Gateway.sol
```diff
@@ -128,16 +136,19 @@ contract Gateway is ProtocolVersioned, Owned {
         // check if the stake request does not exists
         require(stakeRequests[msg.sender].beneficiary == address(0));
 
-        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), _amount));
+        uint256 amountUT = TokenConversionLib.calculateUTAmount(_amount, conversionRate, conversionRateDecimals);
+        uint256 stakedAmount = TokenConversionLib.calculateSTAmount(amountUT, conversionRate, conversionRateDecimals);
```
jasonklein: See comments on the new lib functions, but I don't think we need to calculate amountUT, do we? We just need to pass the `_amount` and get in return `stakedAmount`; since `amountUT` is obtained from the lib and then immediately sent back to the lib, it seems like that can all be done in one call rather than two.

*Alternatively* (but this might be a better question for refactoring), does it make sense to leave the two calls as is, but also store `amountUT` in the `stakeRequest` and pass it in the call to `stake` (which would remove the need to call `calculateUTAmount` in `OpenSTValue`)?

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+        uint256 amountUT)
+    {
+        require(amountST > 0, 'Amount that needs to be converted should be greater than zero');
+        require(conversionRate > 0, 'Conversion Rate should be greater than zero');
+
+        amountUT = (amountST.mul(conversionRate)).div(10 ** uint256(conversionRateDecimals));
+        return amountUT;
+    }
+
+    /**
```
jasonklein: I think this function should take `_amount` (as in what is sent to `requestStake`) and internally call `calculateUTAmount`, to save on calls from `Gateway`. Relatedly, I suggest renaming to something like `calculateMaxStake`.

In addition to any changes related to the above, the `@notice` should not refer to unstaking.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 amountST)
+    {
+        require(amountUT > 0, 'Amount that needs to be converted should be greater than zero');
+        require(conversionRate > 0, 'Conversion Rate should be greater than zero');
+
+        amountST = (amountUT.mul(10 ** uint256(conversionRateDecimals))).div(conversionRate);
```
jasonklein: IIRC, Solidity division always goes to *floor*; but I believe the conclusion we reached is that we round up, so that the amount _staked_ might be max 1 OST Wei higher than the amount minted (hence the loss). Please correct me if I'm mistaken.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
+pragma solidity 0.4.23;
+
+import "./SafeMath.sol";
+
+library TokenConversionLib {
```
jasonklein: We do not typically name our libraries, `...Lib`, so I suggest striking that from the name of this library.

And I suggest raising it as a question for Pro/the style guide, since we do name things `...Interface` and `...Abstract`.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
+pragma solidity 0.4.23;
```
jasonklein: Please add a header consistent with the headers in our other Solidity files.

contracts/TokenConversionLibTest.sol
```diff
@@ -0,0 +1,51 @@
+pragma solidity 0.4.23;
```
jasonklein: Please see header and naming comments for `TokenConversionLib.sol`.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+
+import "./SafeMath.sol";
+
+library TokenConversionLib {
+    using SafeMath for uint256;
+
+    /**
```
jasonklein: Suggest making the documentation of this function less specific, because the question of converted amount need not be limited to the staking context. It's just: given an amount of ST and the factors of a conversion rate, this function returns what the UT into which that amount of ST converts.

Please make "utility" in "Calculate Utility token..." lowercase and please update this documentation (and the documentation across this file) to be consistent with the new style guide (i.e., start with an uppercase letter, except in the case of proper nouns that are lowercased (e.g., variable names), and end with a period).

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+    function calculateUTAmount(
+        uint256 amountST,
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 amountUT)
+    {
+        require(amountST > 0, 'Amount that needs to be converted should be greater than zero');
```
jasonklein: I don't think this require is needed—0 converts into 0, which seems to me an acceptable conclusion.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+        require(conversionRate > 0, 'Conversion Rate should be greater than zero');
```
jasonklein: Similarly to above, I don't think this require is needed—a 0 conversion rate converts an amount to 0. Practically, a 0 conversion rate might not make sense, but I don't know that this library should express an opinion on that.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+
+        amountUT = (amountST.mul(conversionRate)).div(10 ** uint256(conversionRateDecimals));
```
jasonklein: Do you need to set a variable, `amountUT` and return it? Couldn't you simply return `amountST.mul(conversionRate)).div(10 ** uint256(conversionRateDecimals)`?

The question above is less a suggestion of something _to do_ and more a suggestion of something to ask Pro about re: style guide. The question should also include whether we require _naming_ returns (particularly in this kind of library context).

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,59 @@
...
+    function calculateSTAmount(
+        uint256 amountUT,
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 amountST)
+    {
+        require(amountUT > 0, 'Amount that needs to be converted should be greater than zero');
```
jasonklein: Similarly to above, I don't think this require is needed.

contracts/Gateway.sol
```diff
@@ -66,6 +67,9 @@ contract Gateway is ProtocolVersioned, Owned {
     uint256 public bounty;
     /** Storing utility token UUID */
     bytes32 public uuid;
+    /** Token conversion Rate and Decimal example for 1ST = 3.5UT, conversion rate is 35 and conversion decimal is 1 */
```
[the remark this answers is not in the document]
sarvesh-ost: Done  👍

contracts/Gateway.sol
sarvesh-ost: Done 👍  
However, I wanted to discuss the point related to 'loss-less'.

contracts/Gateway.sol
sarvesh-ost: 1. Are you suggesting to create the third method in the library which internally calculates the stake amount(using these two methods)?  These two methods are required, as they are used in OpenST-Value as well.  



2. The second point about refactoring can be taken care with 0.9.4 Gateway development. Please let me know if you have another opinion.

contracts/TokenConversionLib.sol
sarvesh-ost: 1. As mentioned above, let discuss about this.  

2. Done 👍

contracts/TokenConversionLib.sol
sarvesh-ost: Yes, Max loss during stake is 1 wei. 👍

contracts/TokenConversionLib.sol
sarvesh-ost: 1. Done, Removed stake and unstake terminology. Also changing ST to VT(value token)

2. Done 👍

contracts/Gateway.sol
```diff
@@ -101,43 +107,47 @@ contract Gateway is ProtocolVersioned, Owned {
...
+        conversionRateDecimals = _conversionRateDecimals;
     }
 
     /**
      *  @notice External function requestStake.
      *
-     *  @dev In order to request stake the staker needs to approve Gateway contract for stake amount.
-     *       Staked amount is transferred from staker address to Gateway contract.
+     *  @dev In order to request stake the staker needs to approve Gateway contract for request stake amount.
+     *       Maximum possible staked amount is transferred from staker address to Gateway contract given the conversion rate.
```
schemar: Please change to
> Given the conversion rate and the requested stake amount as an upper limit, the maximum stakeable amount is calculated transferred.
It calculates the stakeable amount so that loss of funds is minimised.

Or who has a better idea? Cc @pgev @jasonklein

contracts/Gateway.sol
```diff
@@ -40,7 +41,7 @@ contract Gateway is ProtocolVersioned, Owned {
     /** Events */
 
     /** Below event is emitted after successful execution of requestStake */
-    event StakeRequested(address _staker, uint256 _amount, address _beneficiary);
+    event StakeRequested(address _staker, uint256 _amount, uint256 _stakedAmount, address _beneficiary);
```
schemar: What is the difference between `_amount` and `_stakedAmount`? It would be clearer if `_amount` had a more more descriptive name like `_requestedAmount`.

contracts/Gateway.sol
```diff
@@ -90,6 +94,8 @@ contract Gateway is ProtocolVersioned, Owned {
         WorkersInterface _workers,
         uint256 _bounty,
         bytes32 _uuid,
+        uint256 _conversionRate,
+        uint8 _conversionRateDecimals,
```
schemar: The new parameters are missing from the doc block.

contracts/Gateway.sol
```diff
@@ -90,6 +94,8 @@ contract Gateway is ProtocolVersioned, Owned {
...
         address _openSTProtocol)
```
schemar: All the code we touch now should probably adhere to the [Solidity Style Guide](https://github.com/OpenSTFoundation/openst-protocol/blob/develop/SOLIDITY_STYLE_GUIDE.md#function-declarations). Please check how function declarations should be formatted and update all functions.

contracts/Gateway.sol
```diff
@@ -101,43 +107,47 @@ contract Gateway is ProtocolVersioned, Owned {
...
     /**
      *  @notice External function requestStake.
      *
-     *  @dev In order to request stake the staker needs to approve Gateway contract for stake amount.
-     *       Staked amount is transferred from staker address to Gateway contract.
+     *  @dev In order to request stake the staker needs to approve Gateway contract for request stake amount.
+     *       Maximum possible staked amount is transferred from staker address to Gateway contract given the conversion rate.
      *
-     *  @param _amount Staking amount.
+     *  @param _requestedAmount requested Staking amount.
```
schemar: Why is "Staking" capitalised? Please change according to other comments.

contracts/Gateway.sol
```diff
@@ -101,43 +107,47 @@ contract Gateway is ProtocolVersioned, Owned {
...
 
-        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), _amount));
+        uint256 amountUT = TokenConversionLib.calculateUTAmount(_requestedAmount, conversionRate, conversionRateDecimals);
+        uint256 stakedAmount = TokenConversionLib.calculateVTAmount(amountUT, conversionRate, conversionRateDecimals);
+
+        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), stakedAmount));
 
         stakeRequests[msg.sender] = StakeRequest({
-            amount: _amount,
+            amount : stakedAmount,
```
schemar: Please remove extra space before the colon.

contracts/Gateway.sol
```diff
@@ -101,43 +107,47 @@ contract Gateway is ProtocolVersioned, Owned {
         workers = _workers;
         bounty = _bounty;
         uuid = _uuid;
-
+        conversionRate = _conversionRate;
```
schemar: Should we check here that the conversion rate is not `0`? If it is we will not be able to call the TokenConversionLib.

contracts/TokenConversionLib.sol
schemar: Please fix for all files.

contracts/TokenConversionLib.sol
schemar: Please remove the `Lib` postfix.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
+pragma solidity 0.4.23;
+
+import "./SafeMath.sol";
+
+library TokenConversionLib {
+    using SafeMath for uint256;
+
+    /**
+     *	@notice Calculate Utility token amount based on value token amount, conversion rate and conversion decimal
```
schemar: Please make "utility" in "Calculate Utility token..." lowercase. Please make sure you end all sentences with a period as per our style guide.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
...
+     *
+     *	@param amount of value token needs to be converted to UT.
```
schemar: Rathere "should be" than "needs to be".

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
...
+     *	@param conversionRate rate which is used to convert value token to utility token.
+     *	@param conversionRateDecimals represents number of decimal in conversion rate.
+     *
+     *	@return amountUT amount of utility Token which is converted from value token given the conversion rate.
```
schemar: "Token" should be lower case.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
...
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 amountUT)
+    {
+        require(conversionRate > 0, 'Conversion Rate should be greater than zero');
+
+        amountUT = (amount.mul(conversionRate)).div(10 ** uint256(conversionRateDecimals));
```
schemar: See also below: how does rounding work?

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
...
+        return amountUT;
+    }
+
+    /**
+      *	@notice Calculate Value token amount based on utility token amount, conversion rate and conversion decimal.
```
schemar: "Value" should be lower case.

contracts/TokenConversionLib.sol
```diff
@@ -0,0 +1,57 @@
...
+      */
+
+    function calculateVTAmount(
+        uint256 amount,
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 amountVT)
```
schemar: Please fix function comment to be in line with the style guide (**all** functions).

contracts/TokenConversionLibTest.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import "./TokenConversionLib.sol";
+
+library TokenConversionLibTest {
```
schemar: This file does not include any tests, why is it called `...Test`? What's our convention here? @pgev?

test/TokenConversionLibTest.js
```diff
@@ -0,0 +1,160 @@
...
+
+  let contract;
+
+  describe('ST to UT conversion Test', async () => {
+
+    before(async () => {
+      contract = await TokenConversionLibTest.new();
+    });
+
+    it('should convert ST to UT for 1ST = 1BT', async function () {
```
schemar: It should convert "VT to UT", right?

test/TokenConversionLibTest.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+
+
+    it('should fail if amount that needs to converted is zero', async function () {
+
+      let amount = 0
+        , conversionRate = 10
+        , conversionDecimal = 1;
+
+      utils.expectThrow(contract.calculateUTAmount.call(amount, conversionRate, conversionDecimal));
```
schemar: Does this still hold?

test/TokenConversionLibTest.js
```diff
@@ -0,0 +1,160 @@
...
+      assert.equal(result.toNumber(), 10);
+    });
+
+    it('should fail if amount that needs to converted is zero', async function () {
+
+      let amount = 0
+        , conversionRate = 10
+        , conversionDecimal = 1;
+
+      utils.expectThrow(contract.calculateVTAmount.call(amount, conversionRate, conversionDecimal));
```
schemar: Does this still hold?

contracts/Gateway.sol
```diff
@@ -67,7 +67,8 @@ contract Gateway is ProtocolVersioned, Owned {
     uint256 public bounty;
     /** Storing utility token UUID */
     bytes32 public uuid;
-    /** Token conversion rate and decimal example for 1ST = 3.5UT, conversion rate is 35 and conversion decimal is 1 */
+    /** Token conversion rate and decimal
+    example for 1ST = 3.5UT, conversion rate is 35 and conversion decimal is 1. */
```
schemar: I don't think this formatting is legal with the current style guide. Please read the style guide and adapt if necessary.

contracts/Gateway.sol
```diff
@@ -66,6 +67,11 @@ contract Gateway is ProtocolVersioned, Owned {
     uint256 public bounty;
     /** Storing utility token UUID */
     bytes32 public uuid;
+    /** Token conversion rate and decimal.
```
schemar: Multi line comments should start below the line with the two asterisks. Please see [here](https://github.com/OpenSTFoundation/openst-protocol/blob/develop/SOLIDITY_STYLE_GUIDE.md#multi-line) for examples.

contracts/Gateway.sol
```diff
@@ -84,60 +90,72 @@ contract Gateway is ProtocolVersioned, Owned {
...
-        require(_amount > uint256(0));
+        require(_requestedAmount > uint256(0));
         require(_beneficiary != address(0));
 
         // check if the stake request does not exists
         require(stakeRequests[msg.sender].beneficiary == address(0));
 
-        require(OpenSTValueInterface(openSTProtocol).valueToken().transferFrom(msg.sender, address(this), _amount));
+        //To calculate maximum stakeable amount, UT amount is first calculated with requested stake amount.
+        //This calculated UT amount is converted back to value token which used for staking process.
```
schemar: Please put a single space after `//`, see [here](https://github.com/OpenSTFoundation/openst-protocol/blob/develop/SOLIDITY_STYLE_GUIDE.md#comments)

contracts/Gateway.sol
```diff
@@ -84,60 +90,72 @@ contract Gateway is ProtocolVersioned, Owned {
...
+        conversionRateDecimals = _conversionRateDecimals;
     }
 
     /**
      *  @notice External function requestStake.
      *
-     *  @dev In order to request stake the staker needs to approve Gateway contract for stake amount.
-     *       Staked amount is transferred from staker address to Gateway contract.
+     *  @dev Prior to calling requestStake, the staker must approve the transfer of the requested stake amount to the Gateway.
+     *       Given the conversion rate and the requested stake amount as an upper limit, the maximum stakeable amount is calculated transferred.
```
schemar: Sorry, no "transferred" at the end of the sentence.

contracts/TokenConversion.sol
```diff
@@ -0,0 +1,55 @@
...
+     *	@return amountVT Amount of value tokens which is converted from utility token given the conversion rate.
+     */
+    function calculateVTAmount(
+        uint256 amount,
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        internal
+        pure
+        returns(
+        uint256 /*amountVT*/)
```
schemar: This is not in line with the style guide. Please make sure all your functions adhere to the style guide as shown [here](https://github.com/OpenSTFoundation/openst-protocol/blob/develop/SOLIDITY_STYLE_GUIDE.md#function-declarations)

The closing bracket after the arguments should be on its own line. Same if you have multi-line `return`, which is not required for a single argument.

Example:
```solidity
    function calculateVTAmount(
        uint256 amount,
        uint256 conversionRate,
        uint8 conversionRateDecimals
    )
        internal
        pure
        returns(uint256 /*amountVT*/)
    {
```

Please fix the functions of this PR accordingly.

contracts/TokenConversionTest.sol
```diff
@@ -0,0 +1,49 @@
...
+     */
+    function calculateUTAmount(
+        uint256 amount,
+        uint256 conversionRate,
+        uint8 conversionRateDecimals)
+        public
+        pure
+        returns(
+        uint256)
+    {
```
schemar: Please fix the functions' style in this file.

schemar:
Note: there could be an issue with rounding as previously pointed out by @jasonklein .
@sarvesh-ost will investigate what the consequences could be and whether we need to functions to calculate VT from UT (one for staking and one for redeeming).

benjaminbollen:
@sarvesh-ost should we close this PR and open a new one in token-contracts for branded token later?

Issue #217, opened 2018-08-01
Update solidity style guide


pgev:
Specify:

- Library contracts' naming convention.
- Named return parameters' naming convention.
- Documentation of multiple named return parameters.
- Usages for named and unnamed return parameters.

PR #218, opened 2018-08-01
Update solidty style guide (#217)

pgev:


SOLIDITY_STYLE_GUIDE.md
```diff
@@ -1485,6 +1574,40 @@ function approve(address _spender, uint256 _value) returns (bool success);
...
+/**
+ * Calculates sum and production.
+ *
+ * @param _a First input to calculate sum and production.
+ * @param _b Second input to calculate sum and production.
+ *
+ * @return sum_     Sum of the input arguments.
+ *                  It's really sum of the input arguments.
+ *         product_ Product of the input arguments.
+ *                  It's really product of the input arguments.
```
schemar: It could possibly be required to use a `@return` per return variable, like there is one `@param` per argument. Not sure, though. The only thing I found was [this](https://www.bitdegree.org/learn/solidity-comments/) and I have no idea if it is trustworthy.

pgev: Actually it's not required. There is nothing related to that one in natspec documentation.
I subjectively did not like multiple return tags in documentation, that's why the proposal is to stick to this one for now and update once it's fixed in natspec documentation.

schemar: 👍

Issue #227, opened 2018-08-08
Create unit tests for uncovered lines in Workers.sol


noslav:
* `Workers.sol` has the following coverage report 
60% Statements 9/15 
37.5% Branches 3/8 
50% Functions 3/6 
56.25% Lines 9/16
* Uncovered functions 
removeWorker, remove, approve

noslav:
New report post adding new test cases on Workers.sol -
100% Statements 15/15
50% Branches 4/8
100% Functions 6/6
100% Lines 16/16
All functions are now covered.

schemar:
Dropped.

PR #228, opened 2018-08-09
Unit tests for uncovered lines in Workers.sol

noslav:
Issue ref: #227


benjaminbollen:
these tests for Workers.sol should be tested in (Workers.js which hasn't be copied from openst-payments);

Issue #230, opened 2018-08-09
Create unit tests for uncovered lines in RLP.sol


noslav:
* `RLP.sol` has the following coverage report 
62.93% Statements 73/116
45.95% Branches 34/74
61.54% Functions 16/26
61.15% Lines 96/157
* Uncovered functions 
next, toRLPItem, isNull, isList(117), isData(133), isEmpty, items(167), toAscii, toBool, toByte, toInt, toAddress, _payloadOffset(382, 389, 393), validate

schemar:
Dropped.

PR #232, opened 2018-08-10
Unit tests for uncovered lines RLP.sol 

noslav:
Issue ref: #230


benjaminbollen:
we haven't tested the RLP encoding and decoding in any of these tests

Issue #231, opened 2018-08-09
Create unit tests for uncovered lines in BytesLib.sol


noslav:
* `BytesLib.sol` has the following coverage report 
18.75% Statements 3/16 
0% Branches 0/6 
28.57% Functions 2/7 
20% Lines 5/25
* Functions remaining to be covered 
slice, toBytes32, fromBytes32, toUint, toAddress

noslav:
* Removed 2 redundant fns and added 6 new test cases 
* Solcover report now stands at -
100% Statements 9/9
100% Branches 6/6
100% Functions 5/5
100% Lines 17/17

schemar:
Dropped.

PR #235, opened 2018-08-13
Unit tests for uncovered lines in BytesLib.sol

noslav:
Issue ref: #231


benjaminbollen:
this should be tested in solidity where the type conversion is happening; testing it in js is just adding layers of indirection

PR #239, opened 2018-08-14
Added an M-Contract to register chains

### LINKED ISSUE #238: Register auxiliary chain
As a user
I want to register an auxiliary chain in the mosaic contract
Because I want a dApp to interact with the auxiliary chain

The mosaic contract should automatically give a new chain ID to the new chain. See also [EIP-155](
).
The mosaic contract should make sure any given chain ID is only used once.

schemar:
The Mosaic Contract tracks auxiliary chains and the addresses of their
Core contracts on the origin chain.

There is an open TODO in the `registerChain()` method that I mould like
to get input on.

Fixes #238

### DIFF
--- a/contracts/core/Mosaic.sol
+++ b/contracts/core/Mosaic.sol
@@ -90,7 +90,11 @@ contract Mosaic {
         external
         returns (uint256 chainId_)
     {
-        // TODO: should the contract require that a core contract exists at the given address?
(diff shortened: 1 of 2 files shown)


schemar:
Closing this as we decided that we don’t need a mosaic registry for the time being.

Issue #229, opened 2018-08-09
Create unit tests for uncovered lines in RLPencode.sol


noslav:
* `RLPEncode.sol` has the following coverage report 
24.14% Statements 7/29 
25% Branches 2/8 
40% Functions 2/5 
24.24% Lines 8/33
* Uncovered functions 
encodeItem, encodeList, encodeListWithPasses, encodeLength, getLengthBytes

schemar:
Dropped.

PR #240, opened 2018-08-14
Unit tests for uncovered lines in RLPEncode.sol

noslav:
Issue ref: #229


benjaminbollen:
it is wrong to remove those lines of code from `RLPencode.sol`

Issue #241, opened 2018-08-14
Create unit tests for uncovered lines in MerklePatriciaProof.sol


noslav:
* `MerklePatriciaProof.sol` has the following coverage report 
46.43% Statements 52/112 
38% Branches 19/50 
83.33% Functions 5/6 
54.55% Lines 54/99
* Uncovered functions 
verify(43-46; 69-71), verifyDebug

noslav:
Report currently stands at -
80.36% Statements 90/112
70% Branches 35/50
100% Functions 6/6
84.85% Lines 84/99

schemar:
Dropped.

PR #243, opened 2018-08-14
Unit tests for uncovered lines in MerklePatriciaProof.sol

noslav:
Issue ref: #241


benjaminbollen:
I don't understand why the `verifyDebug` is brought into the main code path.

PR #255, opened 2018-08-16
Report a block to origin

### LINKED ISSUE #253: It should be possible to report an OSTblock
As a validator
I want to report an OSTblock on origin
Because I want to finalise auxiliary's state.

schemar:
Added the ability to report blocks to origin

See TODOs inline and please discuss.

The core contract is now actually tracking reported block headers at
heights. Different block headers can be reported at the same height, but
one header can only be reported once per height. A validator should
probably vote instead if the block has been reported before.

An event informs that a block was reported.

There is also a function for the validator to get all block hashes at a
given height. The validator can use that to check if the block has been
reported before reporting it. If multiple validators report the same
block at the same origin chain height, only one will be the reporter and
pay for the report. The others' transactions will be reverted.

I reduced the interface as some methods are not possible to be
implemented without committing blocks. They should be added when the
voting/committing is implemented.

I cleaned up the syntax and documentation.

I added a new check in the events decoder that will skip any event that
was not emitted by the contract that was passed to the decoder. This
change does not break any existing tests. The test that uses the event
decoder most likely is only interested by events that were emitted by
the provided contract.

Fixes #253


contracts/core/OriginCore.sol
```diff
@@ -0,0 +1,284 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import "../OstInterface.sol";
+import "./OriginCoreConfig.sol";
+import "./OriginCoreInterface.sol";
+
+/**
+ * @title OriginCore is a proof of stake blockchain on Ethereum.
```
benjaminbollen: language update "OriginCore is a meta-blockchain with staked validators on Ethereum."

contracts/core/OriginCore.sol
```diff
@@ -0,0 +1,284 @@
...
+     *      a majority vote from the validators.
+     *
+     * @return The height of the latest committed OSTblock.
+     */
+    function latestBlockHeight()
+        external
+        view
+        returns (uint256)
+    {
+        return height - 1;
```
benjaminbollen: please add a small inline comment explaining why it is `- 1`

contracts/core/OriginCore.sol
```diff
@@ -0,0 +1,284 @@
...
+        uint256 indexed height,
+        bytes32 indexed blockHash
+    );
+
+    /* Structs */
+
+    // TODO: We want to find a way to store the transaction root.
+    /** The header of an OSTblock. */
+    struct Header {
+        // TODO: is recomputation better than storing the header?
```
benjaminbollen: if `.hashed` is only accessed in the `view`able function then it should definitely be calculated on the fly, as the gas is not deducted from the caller.  If later we should require it in a function where we do change state, then we should still default first to calculating it if it is not known from another source - main argument that the only expected hash in the header is the parent hash.

we will introduce the parenthash when we write the commit logic

test/core/OriginCore.js
```diff
@@ -0,0 +1,263 @@
...
+
+            await originCore.reportBlock(
+                expectedBlockHash,
+                blockHeight,
+                0,
+                '0x82ba3527d2433d8a6c50502f341f165955585b1337512e5cd9550eef623c8fd0',
+                '0x82ba3527d2433d8a6c50502f341f165955585b1337512e5cd9550eef623c8fd0'
+            );
+
+            let reportedHashs = await originCore.getReportedBlockHashes.call(blockHeight);
```
benjaminbollen: we can check the event from `reportBlock`

contracts/core/OriginCore.sol
```diff
@@ -0,0 +1,284 @@
...
+        returns (bytes32)
+    {
+        /*
+         * The list of excluded validators is expensive to hash and constant
+         * for all reported blocks at a given height so we omit it from the
+         * block hash definition.
+         */
+        return keccak256(
+            abi.encodePacked(
+                _header.height,
```
benjaminbollen: we should introduce a typeHash (EIP-712)

test/core/OriginCore.js
benjaminbollen: we should assert the block is stored in the mapping `blocks`

test/core/OriginCore.js
```diff
@@ -0,0 +1,263 @@
...
+            );
+            await originCore.reportBlock(
+                expectedBlockHashTwo,
+                blockHeight,
+                0,
+                '0x12345627d2433d8a6c50502f341f165955585b1337512e5cd9550eef62312345',
+                '0x12345627d2433d8a6c50502f341f165955585b1337512e5cd9550eef62312345'
+            );
+
+            let reportedHashes = await originCore.getReportedBlockHashes.call(blockHeight);
```
benjaminbollen: you can query that both blocks are stored in `blocks`

test/core/OriginCore.js
```diff
@@ -0,0 +1,263 @@
...
+
+        it('should reject a report with the wrong hash', async () => {
+            let invalidBlockHash = '0x0000000000000000000000000000000000000000000000000000000000000000';
+            let blockHeight = 0;
+
+            await ost.approve(originCore.address, 10 ** 18);
+
+            let hadError = false;
+            try {
+                await originCore.reportBlock(
```
benjaminbollen: can we not use the `ExpectThrow`?

test/core/OriginCore.js
```diff
@@ -0,0 +1,263 @@
...
+    describe('reporting a block', async () => {
+        let ost;
+        let originCore;
+
+        beforeEach(async () => {
+            ost = await MockToken.new();
+            originCore = await OriginCore.new(3, ost.address);
+        });
+
+        it('should accept a correct block report', async () => {
```
benjaminbollen: please also assert that the balance of the core contract is increased with `COST_REPORT_BLOCK`

test/core/OriginCore.js
schemar: The event won't assert that the report has been stored.

test/core/OriginCore.js
schemar: We cannot as it also succeeds on "out of gas" and "invalid opcode". Instead I will add an `expectRevert`.

PR #259, opened 2018-08-17
An origin state root can be reported to auxiliary

### LINKED ISSUE #258: It should be possible to report an origin block to Auxiliary
As a validator
I want to report an origin block to auxiliary
Because I want to know about the state of origin.

E.g. to open a new block.

schemar:
An AuxiliaryCore contract tracks state roots of Ethereum at their
respective heights.
The cost to report a state root must be paid in the base token of the
auxiliary chain. This should prevent noise in the reports.

A state root can only be reported once per height.

Fixes #258

### DIFF
--- a/contracts/core/AuxiliaryCore.sol
+++ b/contracts/core/AuxiliaryCore.sol
@@ -33,7 +33,7 @@ contract AuxiliaryCore is AuxiliaryCoreInterface, AuxiliaryCoreConfig {
     /* Public Variables */
 
     /** The id of the origin chain that this core tracks. */
-    uint256 public chainIdRemote;
+    uint256 public chainIdOrigin;
 
     /**
      * Maps heights on the Ethereum blockchain to their respoctive reported
@@ -48,9 +48,9 @@ contract AuxiliaryCore is AuxiliaryCoreInterface, AuxiliaryCoreConfig {
 
     /* Constructor */
 
-    /** @param _chainIdRemote The id of the tracked Ethereum chain. */
-    constructor (uint256 _chainIdRemote) public {
-        chainIdRemote = _chainIdRemote;
+    /** @param _chainIdOrigin The id of the tracked Ethereum chain. */
+    constructor (uint256 _chainIdOrigin) public {
+        chainIdOrigin = _chainIdOrigin;
     }
 
     /* External Functions */
--- a/contracts/core/AuxiliaryCore.sol
+++ b/contracts/core/AuxiliaryCore.sol
@@ -1,0 +1,144 @@
+pragma solidity ^0.4.23;
+
+// Copyright 2018 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


schemar:
Also, it is probably a good idea to make the same change here that we did on `OriginCore`, where we track a mapping of the hash to the state root struct, if we can assume that a collision is unlikely.

contracts/core/AuxiliaryCore.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * @notice Returns all the state roots that have been reported at a given
+     *         height.
+     *
+     * @param _height The height for which to get the reported state roots.
+     *
+     * @return The state roots that have been reported at the given height.
+     */
+    function getReportedStateRoots(
```
benjaminbollen: dependent on restructure of `originBlocks`

contracts/core/AuxiliaryCore.sol
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Maps heights on the Ethereum blockchain to their respoctive reported
+     * state root. As Ethereum can fork, multiple state roots can be reported
+     * for the same height.
+     * It is sufficient to only store the state roots of Ethereum. The Casper
+     * FFG rules will assert that only the correct state roots will be
+     * finalised. Falsely reported state roots will consume `msg.value` that
+     * won't ever be returned to the sender of the wrong state root.
+     */
+    mapping (uint256 => bytes32[]) public reportedStateRoots;
```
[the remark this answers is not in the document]
benjaminbollen: on `AuxiliaryCore.sol` we want to report all three types of blocks; `OriginBlock`, `AuxiliaryBlock` and `MetaBlock`, so worth renaming `originBlocks` (or `originStateRoots`)

benjaminbollen:
- the value sent must be exact amount
- I would opt for a similar pattern for all three processes (report, vote, commit) for all blocks (origin, aux, meta)

schemar:
@benjaminbollen  I addressed all comments. Please see the changes and the commit message.
The state roots are now stored in `OriginBlock` structs.

schemar:
When I see [this](https://github.com/ethereum/go-ethereum/blob/9bf6bb8f6388f0f8a6fc5603bb4034a8dc75395c/consensus/ethash/consensus.go#L564) correctly, the reward of the miner is part of the state of the block that is mined. Thus, duplicate state roots at the same height should be impossible. Duplicate state roots at different heights astronomically unlikely. Do you see it the same way @benjaminbollen ?

contracts/core/AuxiliaryCore.sol
```diff
@@ -18,33 +18,45 @@ import "./AuxiliaryCoreConfig.sol";
...
+     * the same height.
      * It is sufficient to only store the state roots of Ethereum. The Casper
      * FFG rules will assert that only the correct state roots will be
      * finalised. Falsely reported state roots will consume `msg.value` that
      * won't ever be returned to the sender of the wrong state root.
      */
-    mapping (uint256 => bytes32[]) public reportedStateRoots;
+    struct OriginBlock {
+        uint256 height;
+        bytes32 rootHash;
```
benjaminbollen: `rootHash` is `stateRoot`

PR #269, opened 2018-08-21
Reporting an auxiliary checkpoint to auxiliary

### LINKED ISSUE #265: It should be possible to report an auxiliary checkpoint on auxiliary
As a validator
I want to report an auxiliary checkpoint
Because I want to be able to vote on it.

The auxiliary checkpoint should be reported to the auxiliary 🥋Core contract.

schemar:
It is now possible to report an auxiliary checkpoint to auxiliary. The
report must be at a valid height in relation to the epoch length and a
checkpoint can only be reported once.

An event announces a successful report.

A mapping stores all reports.

The interface and config have been extended accordingly.

Fixes #265


contracts/core/AuxiliaryCore.sol
```diff
@@ -47,6 +53,17 @@ contract AuxiliaryCore is AuxiliaryCoreInterface, AuxiliaryCoreConfig {
         bytes32 stateRoot;
     }
 
+    /**
+     * An auxiliary checkpoint.
+     *
+     * The blockhash is sufficient to identify the checkpoint. The height is
+     * required to know the epoch and enforce slashing conditions.
+     */
+    struct AuxiliaryCheckpoint {
```
benjaminbollen: we want to know `gas` consumed, and report all auxiliary blocks.  We should report the RLP encoded block header, and decode it to obtain the parameters and hash it to obtain the block hash.  We can only report the child for a reported parent (extracted from header) so that we ensure an unbroken chain. When the reported block height is higher than `block.number - 256` we can assert that we are on the actual history.

PR #277, opened 2018-08-22
Constructing the auxiliary stake contract

### LINKED ISSUE #270: Deploy AuxiliaryStake contract
As a client
I want to deploy a staking contract to auxiliary
Because I want to add auxiliary to mosaic.

In short, the list of validators must be an argument to the constructor.

schemar:
The contract is initialised on auxiliary with a starting set of
validators. The starting set consists of addresses and their deposit.
All starting validators are assumed to join the validator set at height
0 and to not have left the set of validators, yet. The stake must be
greater than zero, otherwise it is irrelevant.

The contract also tracks the total stakes at any given height so it is
known what the weight of a validator is at a given height in comparison
to the total stake.

Fixes #270

### DIFF
--- a/contracts/core/AuxiliaryStake.sol
+++ b/contracts/core/AuxiliaryStake.sol
@@ -1,0 +1,164 @@
+pragma solidity ^0.4.23;
+
+// Copyright 2018 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


schemar:
I think that we do not need the ended field as an endHeight of 0 is impossible and could thus be interpereted as "not ended". But I don't know what the best practice is. It would be kind of "magic", which I would be allergic against 😛

benjaminbollen:
We'll keep actively reworking this to complete logic, so merging this first step

PR #281, opened 2018-08-23
Split AuxiliaryCore into two separate Observers

### LINKED ISSUE #280: Auxiliary core should be split into origin observer and auxiliary observer

schemar:
I split the AuxiliaryCore into an AuxiliaryObserver and an
OriginObserver. Both should be deployed to auxiliary. The auxiliary
observer accepts auxiliary checkpoint reports, and the origin observer
accepts origin block reports.

I split and rearranged the tests accordingly.

Fixes #280


schemar:
Could we have only one observer implementation and depoly it twice? Once to observe auxiliary and once to observe origin? The code looks mostly duplicated ...

benjaminbollen:
as discussed, perhaps both contract can inherit an `Observer.sol`

schemar:
I'll add an issue to refactor the observers later.

PR #289, opened 2018-08-24
New validators are now added at new OSTblocks

### LINKED ISSUE #282: Validator set should be able to change on the stake contract
The staking contract should track changes in the set of validators.
The OSTblock gateway should be able to update the knowledge about added and/or removed validators and their details.
The OSTblock gateway will get the info from newly reported OSTblocks, which contain added and removed validators in their header.

schemar:
The OSTblock gate contract is allowed to update the OSTblock on the
auxiliary stake contract. It increases the OSTblock height and adds the
new validators.

The address of the OSTblock gate is set on construction and cannot be
changed later.

It is currently not allowed to remove validators. That will be added in
a later step and requires stitching with forward and rear validator sets
like described in the Casper FFG paper.

Refactored adding of new validators into a private method as it is used
when updating an OSTblock as well as on construction.

Added JavaScript tests to assert the correct behaviour when initialising
an auxiliary stake contract.

Added Solidity tests to test the actual update method. In order to test
for reverts, I added a revert proxy as described in the truffle
documentation.

Fixes #282

### DIFF
--- a/contracts/core/AuxiliaryStake.sol
+++ b/contracts/core/AuxiliaryStake.sol
@@ -62,6 +62,19 @@ contract AuxiliaryStake {
     /* Public Variables */
 
     /**
+     * The OSTblock gate is the only contract that is allowed to update the
+     * OSTblock height.
+     */
+    address public ostBlockGate;
+
+    /**
+     * Tracks the current height of the OSTblock within this contract. We track
+     * this here to make certain assertions about newly reported OSTblocks and
+     * to know what current height we are voting on.
+     */
+    uint256 public currentOstBlockHeight;
+
+    /**
      * Maps auxiliary addresses of validators to their details.
      *
      * The initial set will be given at construction. Later, validators can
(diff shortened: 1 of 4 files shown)


contracts/core/AuxiliaryStake.sol
```diff
@@ -61,6 +61,19 @@ contract AuxiliaryStake {
...
+
+    /**
+     * Tracks the current height of the OSTblock within this contract. We track
+     * this here to make certain assertions about newly reported OSTblocks and
+     * to know what current height we are voting on.
+     */
+    uint256 public currentOstBlockHeight;
+
     /**
      * Maps auxiliary addresses of validators to their details.
```
benjaminbollen: the validator is a tuple of (origin, auxiliary) key, always referenced by the origin key. can be corrected later

schemar: The thinking was that we only need the auxiliary addresses on auxiliary.
We should never need the addresses from origin here and the auxiliary addresses are unique.
Right?

PR #293, opened 2018-08-27
Message-based gateway

benjaminbollen:


contracts/gateway/Core.sol
```diff
@@ -135,6 +138,21 @@ contract Core is CoreInterface {
...
+	//TODO: This is added for demo purpose.
+	/**
+	 *  @notice The Co-Core address is the address of the core that is
+	 *          deployed on the auxiliary chain. Should only be set if this
+	 *          contract is deployed on the origin chain.
+	 *
+	 *  @param _coCore Address of the Co-Core on auxiliary.
+	 */
+	function setCoCoreAddress(address _coCore) external {
+		require(_coCore != address(0), "Co-Core address is 0");
```
schemar: does it also require `coCore == address(0)`?

contracts/gateway/CoGateway.sol
```diff
@@ -0,0 +1,715 @@
...
+import "./MessageBus.sol";
+import "./CoreInterface.sol";
+import "./EIP20Interface.sol";
+import "./UtilityTokenAbstract.sol";
+
+contract CoGateway {
+
+	using SafeMath for uint256;
+
+	event  StakingIntentConfirmed(
```
schemar: two spaces after `event  `

contracts/gateway/CoGateway.sol
```diff
@@ -0,0 +1,715 @@
...
+		bytes32 messageHash,
+		address staker,
+		uint256 stakerNonce,
+		address beneficiary,
+		uint256 amount,
+		uint256 blockHeight,
+		bytes32 hashLock
```
schemar: indent 8 extra spaces for the events' arguments?

contracts/gateway/CoGateway.sol
```diff
@@ -0,0 +1,715 @@
...
+		address cogateway,
+		address token
+	);
+
+	/* Struct */
+	/**
+	 *  It denotes the redeem request.
+	 *  Status values could be :-
+	 *  0 :- amount used for redemption
+	 *  1 :- beneficiary is the address in the target chain where token will be minted.
```
schemar: I don't see a status that can have a value. Generally, instead of `0` and `1` magic numbers it should use an enum (or possibly constants).

PR #296, opened 2018-08-28
Code clean up

deepesh-kn:
- Deleted old files.
- Renamed V1 files to original names.


contracts/gateway/GatewayRedeemInterface.sol
```diff
@@ -0,0 +1,58 @@
+pragma solidity ^0.4.23;
+
+contract GatewayRedeemInterface {
+
+    function token() external returns (address);
+    function organisation() external returns (address);
```
benjaminbollen: this shouldn't be exposed on an interface.

PR #298, opened 2018-08-28
Created interfaces for the core contracts

### LINKED ISSUE #295: Mosaic interfaces should be defined
The mosaic interfaces for the actors should be defined. This will allow independent implementation of the individual parts.

schemar:
**A thorough review of this PR, preferably by many people, is required. It is the basis for the entire near term core/mosaic implementation.**

Changed some existing interfaces and added some new interfaces. Renamed
all interfaces and contracts to clarify their purpose.

The voter contract is now called "Polling Place". From Wikipedia:

> A polling place is where voters cast their ballots in elections.

There is a single polling place for both block stores. The address of
the block store is an argument to the vote function to clarify what
block store the vote is for.

A parent block store now exists in a rudimentary state. The auxiliary
block store and the origin block store are block stores. The block store
has its own BlockStoreInterface.

All missing methods that exist on the interfaces have been added to the
contracts that implement them with a `revert()` that the method is not
(yet) implemented.

I removed some TODOs and created GitHub issues instead to clean up the
code base.

The stake interface already includes all methods for validators to enter
and leave the validator set. We may want to remove the methods related
to ekiting the set of validators in order to squelch the warnings from
the compiler.

*Open Questions:*

1. Should we reduce the number of arguments to the `vote()` method on
the PollingPlace? An option is a `bytes _voteRlp` argument that is the
entire vote RLP encoded. In the method we would still need to get all
the variables for verification and storage.

2. The `proposeBlock()` method of the `OriginCoreInterface` now has an
argument `_dynasty`. Instead, I think it would make sense to add the
(auxiliary) dynasty to the OSTblock header.

3. Two questions on the vote verification method of the
`OriginCoreInterface`:

    a. How do we link votes (auxiliary FFG) to OSTblocks? Based on the
state root? Then we would need to add the state root to our vote object.
I think a good idea would be to add the block hash of the finalised
auxiliary checkpoint to the OSTblock header.

    b. How are the votes proven? Where is the trie of votes kept? Only in
the rum nodes' memory?

Fixes #295


contracts/core/OriginCore.sol
```diff
@@ -164,6 +163,55 @@ contract OriginCore is OriginCoreInterface, OriginCoreConfig {
...
+     *         it to be committed.
+     *
+     * @param _ostBlockHeaderRlp The entire header of the proposed OSTblock,
+     *                           RLP encoded.
+     * @param _dynasty The dynasty number where the OSTblock closes on the
+     *                 auxiliary chain. It is not part of the OSTblock header
+     *                 and must therefore be presented separately.
+     */
+    function proposeBlock(
+        bytes _ostBlockHeaderRlp,
```
benjaminbollen: for ostBlock we can report the params explicitly

contracts/core/BlockStoreInterface.sol
```diff
@@ -0,0 +1,48 @@
...
+        external
+        returns (bool success_);
+
+    /**
+     * @notice Marks a block in the block store as justified.
+     *         Only the polling place may call this method.
+     *
+     * @param _targetBlockHash The block hash of the block that is justified.
+     */
+    function justify(bytes32 _targetBlockHash) external;
```
benjaminbollen: `justify(bytes32 _source, bytes32 _target)` per our discussion?

contracts/core/OriginCore.sol
```diff
@@ -164,6 +163,55 @@ contract OriginCore is OriginCoreInterface, OriginCoreConfig {
...
+     *                                the last finalised checkpoint to its
+     *                                direct checkpoint. Thus finalising the
+     *                                source of the votes with a supermajority.
+     *
+     * @return `true` if the verification succeeded.
+     */
+    function verifyVote(
+        bytes32 _ostBlockHash,
+        bytes _voteTrieBranchRlp,
+        bytes _childVoteTrieBranchRlp
```
benjaminbollen: not needed

contracts/core/PollingPlaceInterface.sol
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @return `true` if the vote was recorded successfully.
+     */
+    function vote(
+        address _blockStore,
+        bytes32 _source,
+        bytes32 _target,
+        uint256 _sourceHeight,
+        uint256 _targetHeight,
+        bytes32 _hash,
```
benjaminbollen: please remove

schemar:
- Justification now requires source and target
- OSTblock headers are now reported as individual arguments
- Votes are only verified to the child of the finalised checkpoint
- Hash has been removed from the vote interface
- A `_coreId` has been added to the vote interface
    - It is important to note that a list must be maintained externally
that identifies which chain an identifier belongs to.

contracts/core/StakeInterface.sol
```diff
@@ -0,0 +1,151 @@
...
+     * @param _auxiliaryValidatorAddress The address of the validator on
+     *                                   auxiliary.
+     * @param _withdrawalAddress The address where the OST should be
+     *                           transferred in case of a withdrawal.
+     *
+     * @return The unique index of the registered validator.
+     */
+    function deposit(
+        uint256 _amount,
+        address _originValidatorAddress,
```
benjaminbollen: `deposit(uint256 _amount, address _validator)`; `msg.sender` is the owner on origin Core to deposit and withdrawal for the validator address (votes are cast on auxiliary)

contracts/core/StakeInterface.sol
```diff
@@ -0,0 +1,151 @@
...
+     *
+     * @return The full header of the new, opened OSTblock as a result of the
+     *         closing, RLP encoded. The newly opened OSTblock header inculdes
+     *         all changes in the set of validators.
+     */
+    function closeOstBlock(
+        bytes _closedOstBlockHeaderRlp
+    )
+        external
+        returns (bytes _openedOstBlockHeaderRlp);
```
benjaminbollen: `function closeHeight(uint256 _closingHeight) returns (address[] updateValidators_, uint256[] weight_)`

contracts/core/StakeInterface.sol
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * @notice Returns the total stake of all validators at a given height.
+     *
+     * @param _height The height for which to get the total deposit.
+     *
+     * @return The total stake at the given height.
+     */
+    function totalStakeAtHeight(
```
benjaminbollen: `totalWeight(uint256 _height)` only at open height, pass it to assert

contracts/core/StakeInterface.sol
```diff
@@ -0,0 +1,151 @@
...
+     *      reached, therefore committing the OSTblock.
+     *
+     * @param _height The OSTblock height for which to get the stake.
+     * @param _auxiliaryValidatorAddress The address of the validator on the
+     *                                   auxiliary chain.
+     *
+     * @return The stake of the validator at the given height. Can be 0, for
+     *         example when the address never deposited.
+     */
+    function getStakeAtHeightByAuxiliaryAddress(
```
benjaminbollen: `weight(uint256 _height, address _validator)`, only for open height, use height to assert

PR #309, opened 2018-08-31
Block library to decode block headers

### LINKED ISSUE #305: It should be possible to decode block headers
As a smart contract
I want to be able to decode RLP encoded block headers
Because the mosaic vaildators report block headers RLP encoded.

schemar:
A library that converts an RLP-encoded block header into a block header
struct. Can later be extended to full blocks.

Fixes #305


deepesh-kn:
Was wondering if we should pass the rlpEncodedHeader and the all the individual items to the TestBlock.testDecode, so that we can test the failing cases as well ? your thoughts ?

schemar:
Not sure how to test a failing case. You mean if the RLP encoding is incorrect?

schemar:
@deepesh-kn I added two negative test cases.

PR #320, opened 2018-09-05
Class and Sequence Diagrams for Mosaic Core

schemar:
The class (contract) diagram is an overview over the contracts that
mosaic uses and how they relate to each other.

The diagrams were created with draw.io and you can upload the XML files
there to edit the diagrams.

To see the diagrams, check out the PDFs.

There are two sequence diagrams:

1. The overview sequence diagram that shows the entire process
2. A diagram that shows how blocks are reported and voted on in the
block stores. It is refernced from the general diagram.

The second diagram exist once and the `:BlockStore` is not specified as
auxiliary or origin. That is because the process is the same for both
block stores.

I deleted the old diagrams as they are not relevant anymore. They still
exist in the relevant branches and (release) tags.

/cc @achirala


schemar:
Too far away from the paper spec at the moment.

Issue #335, opened 2018-09-25
Votes should be hashed according to EIP-712


schemar:
Transition objects will be hashed accordingly. It will be checked in the block store to deem a vote valid. The vote object should be constructed the same way.

See `MetaBlock` in `contracts/core`.

See also [EIP 712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) and the [example solidity implementation](https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.sol).

Contracts:
- All contracts that work with vote hashes

schemar:
This issue is *only* about vote objects.

PR #399, opened 2018-10-22
Votes should be hashed according to EIP-712 #335

deepesh-kn:
Fixes: #335
- Created `struct VoteMessage`
- Created `struct Vote`
- Added Hashing logic as per EIP-712

Please note:  In this  PR `bytes20 coreIdentifier` is still part of `struct VoteMessage`. 
As the votes are for both origin and auxiliary chain, we need to have the `coreIdentifier` information while generating the `voteHash`.


contracts/core/PollingPlace.sol
```diff
@@ -300,28 +346,27 @@ contract PollingPlace is PollingPlaceInterface {
...
-            _targetHeight
-        );
-        Validator storage validator = getValidatorFromVote(
-            voteHash,
+            _targetHeight,
             _v,
             _r,
-            _s
-        );
+            _s);
```
schemar:
```suggestion
            _s
        );
```

deepesh-kn: Done!

deepesh-kn:
`Did you check whether it is possible to calculate the vote hash only once?`
      Yes, that was missed in the initial commit, I committed that part now, thanks!!!

contracts/core/PollingPlace.sol
```diff
@@ -70,6 +70,52 @@ contract PollingPlace is PollingPlaceInterface {
         uint256 endHeight;
     }
 
+    /** Vote message */
+    struct VoteMessage {
+
+        /** A unique identifier that identifies what chain this vote is about. */
```
schemar: Please also explain here why it has to be part of the struct.

Issue #395, opened 2018-10-22
Stake should not iterate all validators


schemar:
The stake contract iterates over all validators when calculating the total weight at a given height. This was required to know the total weight at all heights without storing all total weights at all heights.

We don't need the total weight at all heights. We only need the total weight at the current height, so that we know when we reached a +2/3 majority on a proposal in order to commit it.

This means that we can have a local variable store the total weight. And update it when a meta-block is committed.

ℹ️ There are three cases right now: (1) a validator gets slashed (effective weight drops to 0), (2) a validator loses weight due to inactivity leak, and (3) a validator logs out.
In case (1), **immediately** reduce the total weight by the previous weight of the validator.
In case (2), reduce the total weight as part of the meta-block commit when the validator inactivity leak is calculated.
In case (3), reduce the total weight as part of the meta-block commit when the validator leaves the set of validators.

Contract: Stake

PR #400, opened 2018-10-22
Stake now has an initial set of validators

### LINKED ISSUE #322: Stake should have an initial set of validators
It is required in order to commit at least the first two meta-blocks.

Contracts:
- Stake

schemar:
Added an initial set of validators to the Stake contract.

The initial set of validators cannot be added upon construction. It is
not possible to transfer token to the contcract during construction, as
`address(this)` is not available. Giving its own address as a
constructor parameter is possible, but error prone.

Instead, a second `initialize()` function was added. It can only be
called once (successfully) and it initializes the initial validator set
at height `0` with the given values.

Furthermore, there exists a new modifier that checks on crucial methods
that the total weight of the validators is above a minimum weight that
is defined during construction.

## Updated tests for the Stake contract
According to the new framework with a directory per contract and a file
per tested method.

Moved shared functions into a utils class.

Replaced bignumber with bn.

Fixes #322


test/core/stake/closeMetaBlock.js
```diff
@@ -0,0 +1,101 @@
...
+        );
+    });
+
+    it('should fail when a wrong height is given', async () => {
+        await Utils.expectFailedAssert(
+            stake.closeMetaBlock(
+                new BN(3),
+                {from: originCoreAccount}
+            )
+        );
```
deepesh-kn: Why we are using `assert` and not `require` in the contract?
Being an external function and `_closingHeight` as param, we can expect wrong height.
We use `assert` when we know things have unexpectedly gone wrong.

contracts/core/Stake.sol
```diff
@@ -341,41 +427,107 @@ contract Stake is StakeInterface {
...
+        uint256 _height
+    )
+        private
+        view
+        returns (uint256 totalWeight_)
+    {
+        for (uint256 i = 0; i < validatorAddresses.length; i++) {
+            address validator = validatorAddresses[i];
+            totalWeight_ += validatorWeightAtHeight(_height, validator);
+        }
```
deepesh-kn: This can potentially create out of gas error.
`totalWeightAtHeight_` is looping throught all the validators, we should store the total weight for each height or for some checkpoint heights so that the iteration will not result in out of gas.

We may consider optimization in future.

test/core/stake/helpers/stake_utils.js
```diff
@@ -0,0 +1,109 @@
...
+        await token.approve(receiver, amount, {from: sender});
+    }
+};
+
+/**
+ * Approves all transfers and then calls initialize on the stake contract.
+ *
+ * @param {StakeInterface} stake The stake contract.
+ * @param {ERC20Interface} token The token used for staking.
+ * @param {address} treasury The address that will send the ERC20 tokens to
```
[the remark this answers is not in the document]
schemar: OK I will add `string`. Not sure what's more helpful to the developer who reads this 🤔

test/core/stake/closeMetaBlock.js
schemar: `closeMetaBlock()` can only be called by the origin core. If the assert fails, there must be a bug in the origin core contract or in the integration of the contracts.

contracts/core/Stake.sol
schemar: Yes you are right. This is addressed in #395

test/core/stake/helpers/stake_utils.js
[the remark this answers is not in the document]
deepesh-kn: I think it is `type`.
So I was in confusion if `StakeInterface `, `ERC20Interface ` also should be `Object`.

@sarvesh-ost , what your think ?

test/core/stake/closeMetaBlock.js
deepesh-kn: We can keep `assert` for now, we can check if it should be `require` or `assert` when origin core implements the calling of `closeMetaBlock`.

contracts/core/Stake.sol
```diff
@@ -341,41 +436,107 @@ contract Stake is StakeInterface {
...
      * @param _validator The address of the new validator.
-     * @param _amount The initial stake of the new validator.
+     * @param _amount The initial deposit of the new validator.
+     * @param _startingHeight The height at which the validator will enter the
+     *                        set of validators.
      *
      * @return The index under which the new validator is stored.
      */
     function registerNewValidator(
+        address _depositor,
```
sarvesh-ost: Why `_depositor ` address is saved in storage? 
will this be used for during `logout` ?

schemar: `address` is a value type, not a complex type. That means by default, solidity puts it on the EVM *stack:* [Relevant solidity FAQ](https://solidity.readthedocs.io/en/v0.4.24/frequently-asked-questions.html#what-is-the-memory-keyword-what-does-it-do).
Value types are always copied when they are passed as function arguments (in memory).

Only complex types (arrays and structs) have the optional data location to be specified as `storage` or `memory`. I cannot specify `address storage` or `address memory`, for example.

Furthermore, the default for function arguments would be `memory`, not `storage`.

Regardless, `_depositor` *should* be checked during `logout`. It is stored in the validator:
```solidity
Validator memory validator = Validator(
    _depositor,
    _validator,
    _amount,
    _startingHeight,
    false,
    0
);

validators[_validator] = validator;
```
And later checked to be the correct one when logging out. Something like (wild guess):
```solidity
require(
    msg.sender == validator.depositor,
    "Message sender must be depositor of validator."
);
```

PR #414, opened 2018-10-25
Auxiliary block store now uses correct transition object

### LINKED ISSUE #336: Transition objects should be merged
`OriginCore` and `AuxiliaryBlockStore` both use the same transition object. It should only be defined in a single place.

schemar:
The auxiliary transition object was not used anywhere in the code. It is
now used wherever we need the auxiliary transition object. Mostly the
auxiliary block store.

In order to be able to calculate the transition object and hash at an
auxiliary checkpoint, we need to know the origin dynasty and latest
finalized block hash. Those two fields are now tracked on the auxiliary
block store per reponted block by calling on the origin block store. The
address of the origin block store was added to the constructor.

To decide whether a vote is valid, the auxiliary block store now
verifies the given transition hash against the hash that it calculates
itself based on the data that is available  at the source checkpoint. To
do that, `isValidTransitionHash` has been made internal on the
`BlockStore` and is overridden by the `AuxiliaryBlockStore`.

Added `getHead()` and `getCurrentDynasty()` to the `BlockStoreInterface`
and `BlockStore` and `BlockStoreMock`. The auxiliary block store uses
these functions to get the latest observations of the origin block store
when it stores a block report.

I updated the auxiliary block store tests to check for the corret
transition hash. Furthermore I added the accumulated values to the
shared data object.

I added solidity tests for the `MetaBlock` library.

All core identifier types were changed to `bytes20` which is the current
implementation. Must be changed to `bytes32` across the whole project
coordinated.

I tried caching only the transition hash per reported block in the
auxiliary block store instead of all the individual data points like
accumulated gas and accumulated transaction root. It turned out
impossible to do, as the dynasty of a block can change when an earlier
block gets finalized. Thus, the transition hash cannot be cached at the
time of a block report.

Fixes #336

### DIFF
--- a/contracts/core/AuxiliaryBlockStore.sol
+++ b/contracts/core/AuxiliaryBlockStore.sol
@@ -15,6 +15,7 @@ pragma solidity ^0.4.23;
 // limitations under the License.
 
 import "../lib/Block.sol";
+import "../lib/MetaBlock.sol";
 import "../lib/SafeMath.sol";
 import "./BlockStore.sol";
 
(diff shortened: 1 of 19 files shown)


schemar:
>Variableshead and currentDynasty are already public.
>Now we have introduced getHead and getCurrentDynasty the variables can be private or internal.
Is there a "best practice" for this 🤔

deepesh-kn:
> > Variableshead and currentDynasty are already public.
> > Now we have introduced getHead and getCurrentDynasty the variables can be private or internal.
> 
> Is there a "best practice" for this 🤔

The compiler automatically creates getter functions for all public state variables. We already have `head` and `currentDynasty` functions.  With `getHead` and `getCurrentDynasty` we have multiple functions for same task. I am not sure if its best practice, but we followed this practice in past https://github.com/OpenSTFoundation/openst-payments/blob/master/contracts/Pricer.sol

schemar:
It potentially saves on the overall byte size of the compiled contract if it would actually add the public getters again even though they exist 🤔 
Will make them private 👍

PR #443, opened 2018-10-26
Updated SafeMath from the latest OpenZeppelin code

### LINKED ISSUE #378: Update SafeMath

schemar:
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/math/SafeMath.sol

Fixes #378


contracts/lib/SafeMath.sol
```diff
@@ -1,6 +1,6 @@
-pragma solidity ^0.4.23;
+pragma solidity ^0.4.24;
```
deepesh-kn: Other solidity files are using 0.4.23. Should we consider using 0.4.24 in all new files?

contracts/lib/SafeMath.sol
```diff
@@ -15,98 +15,128 @@ pragma solidity ^0.4.23;
...
+            c >= a,
+            "Overflow when adding."
+        );
 
         return c;
     }
 
     /**
-    * @notice Divides two numbers and returns the remainder.
+    * @notice Divides two numbers and returns the remainder (unsigned integer
```
deepesh-kn: indentation is incorrect.

contracts/lib/SafeMath.sol
schemar: Hm, I don't know. As solidity is developing relatively quickly including the addition of new features, I would say update whatever you work on? What do you think?

deepesh-kn: We can keep `0.4.23`, and when we plan to upgrade we can create a ticket.

Issue #382, opened 2018-10-15
BlockStore should have a getter for the transition object


schemar:
## User Story
As a validator
I want to get the transition object from a smart contract
Because I don't want to know how it is generated.

## Scenario
Given that I want to propose a meta-block and need the transition object
When I use the getter on the block store
And specify the block hash for which I want the transition object
Then I get the transition object of the given block hash.

## Info
Later we may generate the transition object in the validator node. But for now there should be a getter on the block store.

Contracts:
- BlockStore

schemar:
The transition object is different for the auxiliary block store (auxiliary transition object) and the general block store (origin transition object).

schemar:
We cannot return a struct as it should be callable from web3.

PR #449, opened 2018-11-06
Block store should have a getter for the transition object

sarvesh-ost:
This PR implements following PR 

1. Getter for origin transition object.
2. Getter for auxiliary transition object.
3. Added initial kernel Hash and mapping to store kernel hash against block hash in auxiliary block store. 
4. Unit test cases

Note: Team discussed that `kernelHash` can be part of `CheckPoint` struct but Same `checkpoint`data structure is used to track origin checkpoint which ideally shouldn't know about kernel hash. Hence a mapping is created with `blockHashVsKernelHash`. 

#382


contracts/core/BlockStore.sol
```diff
@@ -427,6 +427,38 @@ contract BlockStore is BlockStoreInterface {
...
+     * @return blockHash_ Hash of the block at checkpoint.
+     */
+    function transitionObjectAtBlock(
+        bytes32 _blockHash
+    )
+        external
+        returns(bytes20 coreIdentifier_, uint256  dynasty_, bytes32 blockHash_)
+    {
+        require(
+            checkpoints[_blockHash].blockHash == _blockHash,
```
schemar: `isCheckpoint()`

test/core/auxiliary_block_store/transition_object.js
```diff
@@ -0,0 +1,122 @@
...
+            initialBlockHash,
+            initialStateRoot,
+            initialHeight,
+            initialGas,
+            initialTransactionRoot,
+            initialKernelHash
+        );
+    });
+
+    it('should return origin transition object at given block Hash if' +
```
schemar:
```suggestion
    it('should return auxiliary transition object at given block Hash if' +
```

test/core/auxiliary_block_store/auxiliary_transition_object_at_block.js
```diff
@@ -0,0 +1,122 @@
...
+            initialBlockHash,
+            initialStateRoot,
+            initialHeight,
+            initialGas,
+            initialTransactionRoot,
+            initialKernelHash
+        );
+    });
+
+    it('should return origin transition object at given block Hash if' +
```
schemar:
```suggestion
    it('should return auxiliary transition object at given block Hash if' +
```

PR #451, opened 2018-11-06
Block store should have a getter for the transition hash.

sarvesh-ost:
⚠️  This should be merged after #449  

1. It implements the getter for origin transition hash.
1. It implements the getter for auxiliary transition hash.


contracts/core/AuxiliaryBlockStore.sol
```diff
@@ -159,6 +169,95 @@ contract AuxiliaryBlockStore is BlockStore {
...
+     * @return transitionHash_ Hash of transition object at the checkpoint.
+     */
+    function  auxiliaryTransitionHashAtBlock(
+        bytes32 _blockHash
+    )
+        external
+        returns(bytes32 transitionHash_)
+    {
+        require(
+            checkpoints[_blockHash].blockHash == _blockHash,
```
schemar: `isCheckpoint()`

contracts/core/AuxiliaryBlockStore.sol
```diff
@@ -159,6 +169,95 @@ contract AuxiliaryBlockStore is BlockStore {
...
+            "Checkpoint not defined for given block hash."
+        );
+
+        transitionHash_ = MetaBlock.hashAuxiliaryTransition(
```
schemar: Would solidity understand `MetaBlock.hashAuxiliaryTransition(auxiliaryTransitionObjectAtBlock(_blockHash))` ? 🤔

contracts/core/BlockStore.sol
```diff
@@ -427,6 +427,68 @@ contract BlockStore is BlockStoreInterface {
...
+     * @return transitionHash_ Hash of the transition object at the checkpoint.
+     */
+    function transitionHashAtBlock(
+        bytes32 _blockHash
+    )
+        external
+        returns(bytes32 transitionHash_)
+    {
+        require(
+            checkpoints[_blockHash].blockHash == _blockHash,
```
schemar: `isCheckpoint()`

contracts/core/BlockStore.sol
```diff
@@ -427,6 +427,68 @@ contract BlockStore is BlockStoreInterface {
...
+            "Checkpoint not defined for given block hash."
+        );
+
+        transitionHash_ = MetaBlock.hashOriginTransition(
+            checkpoints[_blockHash].dynasty,
+            _blockHash,
+            coreIdentifier
```
schemar: Same question as for auxiliary block store.

test/core/auxiliary_block_store/transition_hash.js
```diff
@@ -0,0 +1,103 @@
...
+            transactionRoot:initialTransactionRoot,
+        };
+
+        let expectedTransitionHash = MetaBlockUtils.hashAuxiliaryTransition(auxiliaryTransitionObject);
+
+        let transitionHash = await  blockStore.auxiliaryTransitionHashAtBlock.call(
+            initialBlockHash
+        );
+
+        assert.equal(
```
schemar: `strictEqual`

test/core/auxiliary_block_store/transition_hash.js
```diff
@@ -0,0 +1,103 @@
...
+            initialBlockHash,
+            initialStateRoot,
+            initialHeight,
+            initialGas,
+            initialTransactionRoot,
+            initialKernelHash
+        );
+    });
+
+    it('should return origin transition object at given block Hash if' +
```
schemar:
```suggestion
    it('should return auxiliary transition object at given block Hash if' +
```

contracts/core/AuxiliaryBlockStore.sol
schemar: still open

contracts/core/AuxiliaryBlockStore.sol
schemar: did you try this?

test/core/auxiliary_block_store/auxiliary_transition_hash_at_block.js
```diff
@@ -0,0 +1,103 @@
...
+            initialBlockHash,
+            initialStateRoot,
+            initialHeight,
+            initialGas,
+            initialTransactionRoot,
+            initialKernelHash
+        );
+    });
+
+    it('should return origin transition object at given block Hash if' +
```
schemar:
```suggestion
    it('should return auxiliary transition object at given block Hash if' +
```

contracts/core/AuxiliaryBlockStore.sol
sarvesh-ost: We definetly need to change `auxiliaryTransitionObjectAtBlock` from `external` to `public`

I will check this.

test/core/auxiliary_block_store/transition_hash.js
schemar: Always good to prevent JS weirdness

contracts/core/AuxiliaryBlockStore.sol
schemar: We have to keep it `external` to keep it in the interface as in the future all interface methods must be `external`. The only option would be to add a public method like `auxiliaryTransitionObjectAtBlock_()` that is called from `auxiliaryTransitionObjectAtBlock()`. Don't worry about it for now.

sarvesh-ost: Just for curiosity, I tried changing `external` to `public`. Seems solidity is not smart enough to unwrap `returns` into `arguments`. 
It gives `Wrong argument count for function call: 1 arguments given but expected 8.`

Issue #404, opened 2018-10-23
JavaScript code linting and formatting


schemar:
We should have standards for how JavaScript code is linted and formatted so it is homogeneous across files.

Copy the tools and settings from the openst repository that @pgev established.

Include auto-formatting of all existing files in this ticket.

Change indentation to `2` spaces.
Change line width hard wrap limit max character to `79` spaces.

hobofan:
The basis for this was added with #458. The existing files in the `test` directory still need to be checked though.

PR #458, opened 2018-11-14
Add initial version of genesis deployment tool

hobofan:
Adds the first version of the genesis deployment script(s).

This allows us to generate the `accounts` section for a parity chainspec, taking as input the bytecode of the compiled contracts and their constructor arguments (+ ABI). This greatly simplifies the bootstrapping of a new Auxiliary chain.

Also adds `eslint` for JS style conformance as per #404.

**TODO**:
- [x] Find a way to be able to invoke script directly instead of via `truffle exec` (truffle exec has a poluted STDOUT output)
- [x] A bit simpler interface for adding Truffle contracts
- [x] Add script to merge the output of one of these deployment scripts into a chainspec
- [x] Allow for specifying a address for a contract instance instead of incrementing from a fixed address.
- [x] More robust way of incrementing addresses (that works for more than 255 addresses)
- [x] Better documentation of the `ContractRegistry` functions

@pgev : I gave you access to my fork


tools/deployment_tools.js
```diff
@@ -0,0 +1,180 @@
...
+    this.linkBytecode = this.linkBytecode.bind(this);
+    this.linkedBytecodeForContract = this.linkedBytecodeForContract.bind(this);
+    this.truffleDeployer = this.truffleDeployer.bind(this);
+    this.toParityGenesisAccounts = this.toParityGenesisAccounts.bind(this);
+    this.takeNextAddress = this.takeNextAddress.bind(this);
+  }
+
+  takeNextAddress() {
+    const addressHex = web3.utils.bytesToHex(this.nextAvailableAddress);
+    this.nextAvailableAddress[this.nextAvailableAddress.length - 1] += 1;
```
schemar: Wouldn't it be easier to have a `BN` or similar and actually increase that? If I understand this approach correctly it is bound to 10 addresses, which may or may not be sufficient in the future.

hobofan: It's bound to 255 (as it increments the last byte, not string character), but yeah, probably possible to use `BN` instead. I'll look into that.

schemar: Doesn't it increment the last hex character (4 bits => max 16)? And does it know to go from `9` to `a`? JavaScript magic? 🤔

hobofan: `this.nextAvailableAddress` is a byte array, so no magic there 😉

hobofan: But that code is gone now anyway 😄

schemar: Ah, somehow I thought it was a hex string 😳

hobofan:
From my side this would be mergeable in the current state. Should be in a usable state for @pgev's work on the ValidatorSet. Next step would be applying this to the deployment sequence for auxiliary, now that #448 is merged.

hobofan:
WIP again following some feedback from @pgev

hobofan:
Back to review.

@pgev I added the state machine we talked about. Also added some tests for core parts of the functionality and a referencing system for contract addresses in constructor arguments (so we can also order those correctly).

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+     * Returns the next available address.
+     *
+     * @return {string} Next address that will be used for a contract created by
+     *                  a transaction from `fromAddress` at the current transaction
+     *                  count (= nonce).
+     */
+    generateAddress() {
+        const addressBytes = Web3.utils.sha3(rlp.encode([this.fromAddress, this.nextNonce]))
+            .slice(12)
+            .substring(14);
```
schemar: Should be extracted into a function with a meaningful name or alternatively explain in a comment what/why it does.

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+        this.contracts = [];
+
+        this.addContract = this.addContract.bind(this);
+        this.toParityGenesisAccounts = this.toParityGenesisAccounts.bind(this);
+    }
+
+    /**
+     * Add a contract to the registry.
+     *
+     * @param {Contract} contracts Contract to add to the registry.
```
schemar:
```suggestion
     * @param {Contract} contract Contract to add to the registry.
```

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+    /**
+     * Generate the "accounts" object for a parity chainspec for all the contracts have been added.
+     * For that the { "0x...": { "constructor": "0x..." } } version of account intialization
+     * which is currently exclusive to parity is used.
+     *
+     * See {@link https://wiki.parity.io/Chain-specification} for more details on the
+     * parity chainspec format.
+     *
+     * @param {Object} options.addressGenerator Address generator to use.
+     *                 Defaults to a IncrementingAddressGenerator.
```
schemar:
```suggestion
     *                 Defaults to an IncrementingAddressGenerator.
```

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+        };
+        const mergedOptions = Object.assign(defaultOptions, options);
+
+        // prepare contracts by ordering and generating addresses
+        const { addressGenerator } = mergedOptions;
+        this._orderContracts();
+        this.contracts.forEach(contract => contract.setAddress(addressGenerator));
+        this.contracts.forEach(contract => contract.instantiate());
+
+        const output = {
```
schemar: Better name? `genesisAccounts`?

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+        return output;
+    }
+
+    /**
+     * Generate transaction objects that can be passed as an argument
+     * to `web3.eth.sendTransaction()`.
+     * This allows for deployment on a live network.
+     *
+     * @param {string} fromAddress See {@ IncrementingNonceAddressGenerator#constructor}.
+     * @param {number} startingNonce See {@ IncrementingNonceAddressGenerator#constructor}.
```
schemar: `@link`s?

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+                };
+            });
+
+        return output;
+    }
+
+    /**
+     * Generate transaction objects that can be passed as an argument
+     * to `web3.eth.sendTransaction()`.
+     * This allows for deployment on a live network.
```
schemar: It generates the objects *from this `ContractRegistry`,* right?
Maybe also explain that transactions and the parity genesis account are alternatives.

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+
+        return transactionObjects;
+    }
+
+    /**
+     * Orders all contracts in the order of sequential deployment.
+     * This is determined by the amount of transitive dependencies (including self).
+     */
+    _orderContracts() {
+        this.contracts.sort((a, b) => a._getDependenciesCount() - b._getDependenciesCount());
```
schemar: I don't understand how the dependencies count guarantees the order? Can't a contract with `n` dependencies depend on a contract with `n+1` or `n-1` dependencies? 😕

tools/merge_chainspec_accounts.js
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env node
+// Called like this:
+// node merge_chainspec_accounts.js chainspec.json accounts.json
```
schemar: you should also explain where chainspec and accounts comes from.

tools/merge_chainspec_accounts.js
```diff
@@ -0,0 +1,19 @@
...
+const accounts = JSON.parse(fs.readFileSync(accountsPath));
+
+Object.entries(accounts).forEach(([address, obj]) => {
+  if (chainspec.accounts[address]) {
+    console.warn(`Account at address ${address} already exists.`);
+  }
+  chainspec.accounts[address] = obj;
+});
+
+console.log(JSON.stringify(chainspec, null, 4));
```
schemar: What is the purpose of this?

tools/test/deployment_tools.test.js
```diff
@@ -0,0 +1,164 @@
...
+const ContractRegistry = require('../deployment_tools.js').ContractRegistry;
+const IncrementingNonceAddressGenerator = require('../deployment_tools.js').IncrementingNonceAddressGenerator;
+
+const EIP20TokenFile = require('../../build/contracts/EIP20Token.json');
+
+const rootDir = `${__dirname}/../../`;
+
+describe('Contract', () => {
+    describe('.setAddress()', () => {
+        test('correctly sets a fixed address for contract', () => {
```
schemar: We usually use mocha `it` as per truffle. Is that an option here?

tools/test/deployment_tools.test.js
```diff
@@ -0,0 +1,164 @@
...
+            const EIP20Token = new Contract(EIP20TokenFile.contractName, EIP20TokenFile.bytecode, constructorAbi, ['MYT', 'MyToken', 18]);
+
+            EIP20Token.setAddress('0x0000000000444440000000000000000000000100');
+
+            const registry = new ContractRegistry();
+            registry.addContract(EIP20Token);
+
+            const output = registry.toParityGenesisAccounts();
+            expect(Object.keys(output)).toContain('0x0000000000444440000000000000000000000100');
+            expect(Object.keys(output).length).toEqual(1);
```
schemar: We usually use Chai `assert`s as per truffle. Is that an option here?

tools/test/deployment_tools.test.js
hobofan: We can definitely switch it to mocha + chai 👍 . I just picked Jest as I was most used to it and there wasn't a general JS testing framework in place.

tools/merge_chainspec_accounts.js
hobofan: 👍 
I'll add some docs for that script.

tools/deployment_tools.js
hobofan: It's not just the count of dependencies but transitive dependencies. There is some graph theory behind it that @pgev can probably better explain 😅

tools/deployment_tools.js
hobofan: Yes! 👍

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
+const Web3 = require('web3');
+const fs = require('fs');
+const path = require('path');
+const rlp = require('rlp');
```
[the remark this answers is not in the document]
hobofan: This should be according to JS best practices (which doesn't really prescribe either). The uppercase version of `Web3` is a bit unnormal, but it appears that this something that Ethereum JS community has settled on (as `web3` is used for an instance).

tools/deploy_token_genesis.js
```diff
@@ -0,0 +1,12 @@
+// Demonstrates a genesis deployment of a EIP20Token
+const Contract = require('./deployment_tools.js').Contract;
+const ContractRegistry = require('./deployment_tools.js').ContractRegistry;
+
+const EIP20Token = Contract.loadTruffleContract('EIP20Token', ['MYT', 'MyToken', 18]);
+EIP20Token.address = '0x0000000000444440000000000000000000000100';
+// EIP20Token.instantiate(['MYT', 'MyToken', 18]);
```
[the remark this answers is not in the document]
hobofan: 👍 Will delete. That was my manual test case before I added the tests.

package.json
```diff
@@ -6,9 +6,13 @@
...
     "assert": "1.4.1",
     "bn.js": "4.11.8",
+    "eslint": "^5.9.0",
+    "eslint-config-airbnb-base": "^13.1.0",
+    "eslint-plugin-import": "^2.14.0",
     "ganache-cli": "6.1.8",
+    "jest": "^23.6.0",
     "keccak": "1.4.0",
-    "rlp": "2.1.0",
+    "rlp": "^2.1.0",
```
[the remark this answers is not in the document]
hobofan: 👍 Will do.

Is there a specific reasoning behind it? Now that lockfiles are a thing in npm-land, I don't see much of a benefit there.

tools/deployment_tools.js
schemar: Ahhhh, yes. Now it makes sense 👍 I forgot to take the transitivity into account 😅

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+     *         genesis file for contract deployment.
+     */
+    generateAddress() {
+        const addressHex = this.nextAvailableAddress;
+
+        // Incrementing next available address.
+        const nextAddressBN = Web3.utils.toBN(addressHex).add(Web3.utils.toBN('1'));
+        this.nextAvailableAddress = `0x${Web3.utils.padLeft(nextAddressBN, 40)}`;
+
+        return addressHex;
```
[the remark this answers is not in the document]
hobofan: No, it starts at `0x0000000000000000000000000000000000010000`

- saves address to variable `addressHex` (which is initially at `0x0000000000000000000000000000000000010000`)
- increments the internal variable used to track the next address
- returns `addressHex` which still is at `0x0000000000000000000000000000000000010000`

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+     */
+    static loadTruffleContract(contractName, constructorArgs, options = {}) {
+        const defaultOptions = {
+            rootDir: `${__dirname}/../`,
+        };
+        const mergedOptions = Object.assign(defaultOptions, options);
+
+        const contractPath = path.join(mergedOptions.rootDir, `build/contracts/${contractName}.json`);
+        const contents = fs.readFileSync(contractPath);
+        const truffleJson = JSON.parse(contents);
```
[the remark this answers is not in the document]
hobofan: One good JS idiomatic alternative would be to have `loadTruffleContract` as a free-standing function, but I generally prefer keeping alternative constructors inside the class itself.

tools/deployment_tools.js
```diff
@@ -0,0 +1,499 @@
...
+    /**
+     * Tries to interpret the provided constructor argument as a contract reference,
+     * and returns the referenced contract if it is one. See {@link Contract#reference}.
+     *
+     * @param {*} constructorArg A constructor argument that is possibly a contract reference.
+     *
+     * @returns {Contract|null} The referenced contract.
+     */
+    _getReferenceContract(constructorArg) {
+        if (!((typeof constructorArg === 'object') && (constructorArg !== null)) || constructorArg.__type !== TYPE_CONTRACT_REFERENCE) {
```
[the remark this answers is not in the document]
hobofan: I've tried a few things, but none of them were really an improvement. I'm open to suggestions!

tools/README.md
```diff
@@ -0,0 +1,58 @@
+## Deployment tool
+
+There is a JS library under `tools/deployment_tool`, which helps in building deployment scripts for non-traditional scenarios.
+
+### Features
+
+- Can output the `accounts` section for a [Parity chainspec][parity-chainspec] to allow
```
schemar:
```
 - Can output the `accounts` section for a [Parity chainspec][parity-chainspec] to allow
```
or
```suggestion
- Can output the `accounts` section for a [Parity chainspec](parity-chainspec) to allow
```

(compare line [43](https://github.com/OpenSTFoundation/mosaic-contracts/pull/458/files#diff-07b723ae57aaa00bce72a3022cf09ecdR43))

package.json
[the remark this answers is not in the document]
schemar: In theory, I agree. However, the lock-file in npm behaves differently than what I would expect. For example, it gets changed when I run `npm install` (blows my mind).

Anyways, any update (even a bug-release), should always be a very conscious choice. That's pretty much the only reason.

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,323 @@
...
+    /**
+     * Tries to interpret the provided constructor argument as a contract reference,
+     * and returns the referenced contract if it is one. See {@link Contract#reference}.
+     *
+     * @param {*} constructorArg A constructor argument that is possibly a contract reference.
+     *
+     * @returns {Contract|null} The referenced contract.
+     */
+    _getReferenceContract(constructorArg) {
+        if (!((typeof constructorArg === 'object') && (constructorArg !== null)) || constructorArg.__type !== TYPE_CONTRACT_REFERENCE) {
```
schemar: Maybe something like the below. I am sure this can inspire you to do something readable 😄 
```js
if(!isReference(constructorArg)) {
    return null;
}


function isReference(constructorArg) {
    if (!isInstance(constructorArg) {
        return false;
    }

    return constructorArg.__type === TYPE_CONTRACT_REFERENCE
}

function isInstance(constructorArg) {
    return typeof constructorArg === 'object' && constructorArg !== null;
}

```

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,323 @@
...
+            throw new Error(`Expected constructor arguments for constract ${this.contractName}`);
+        }
+        if (this.constructorArgs && !this.constructorABI) {
+            throw new Error(`Provided arguments for contract ${this.contractName}, but no constructorAbi is set.`);
+        }
+
+        this._linkBytecode();
+
+        let constructorData = this.linkedBytecode();
+        constructorData += this._encodeConstructorArguments();
```
schemar: Mind adding a little bit of documentation to the above two lines? That would really help me a lot in understanding it.

tools/deployment_tool/contract_registry.js
```diff
@@ -0,0 +1,127 @@
...
+                };
+            });
+
+        return genesisAccounts;
+    }
+
+    /**
+     * Generate transaction objects that can be passed as an argument
+     * to `web3.eth.sendTransaction()`.
+     * This allows for deployment on a live network.
```
schemar: It would help to reference `toParityGenesisAccounts` and `toLiveTransactionObjects` to each other I think. What do you think?

tools/README.md
hobofan: Should be []

package.json
```diff
@@ -6,8 +6,13 @@
     "abi-decoder": "1.2.0",
     "assert": "1.4.1",
     "bn.js": "4.11.8",
+    "chai": "4.2.0",
+    "eslint": "5.9.0",
```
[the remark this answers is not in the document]
hobofan: Nope! Updating!

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * Helper for loading a contract saved in a format followin the truffle-contract-schema.
+     *
+     * See {@link https://github.com/trufflesuite/truffle/tree/66e3b1cb10df881d80e2a22ddea68e9dcfbdcdb1/packages/truffle-contract-schema}
+     *
+     * @param {string} contractName Name of the contract to load.
+     * @param {Array<*>} constructorArgs Arguments for the contract constructor.
+     * @param {object} options
```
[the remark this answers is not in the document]
hobofan: Maybe.

tools/deployment_tool/contract.js
hobofan: 👍

tools/deployment_tool/contract_registry.js
hobofan: Done!

hobofan:
:repeat: :)

tools/deployment_tool/contract.js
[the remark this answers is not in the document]
schemar: Then it could as well be a `rootDir` argument 🤷‍♂️

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,323 @@
...
+     */
+    addLinkedDependency(linkedContract) {
+        this._ensureState(STATE_NEW);
+        this.linkReplacements.push(linkedContract);
+    }
+
+    /**
+     * Returns the contract's address as detemined by the provided AddressGenerator.
+     *
+     * @param {Object|string} addressGeneratorOrAddress The AddressGenerator used for
```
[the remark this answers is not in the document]
schemar: Someone else should probably comment on this @benjaminbollen @pgev @deepesh-kn

tools/merge_chainspec_accounts.js
schemar: I don't know enough about JS, but are you using a *logging* mechanism to give feedback to the user on `stdout`? Can I change the behavior of `console.log()` to log somewhere else? E.g. a file? In that case I would prefer explicitly writing to `stdout` and not a log.

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,342 @@
...
+const STATE_INSTANTIATED = 'instantiated';
+
+const TYPE_CONTRACT_REFERENCE = 'contract_reference';
+
+class Contract {
+    /**
+     * @param {string} contractName Name of the contract to create. If this contract is
+     *                 linked into another contracted, the provided contractName has to
+     *                 match the linking placeholder.
+     * @param {string} contracBytecode Bytecode of the contract.
```
deepesh-kn: Typo: `contracBytecode` -> `contractBytecode`

tools/deployment_tool/contract.js
```diff
@@ -0,0 +1,342 @@
...
+    /**
+     * Add a linked dependency to the contract.
+     */
+    addLinkedDependency(linkedContract) {
+        this._ensureState(STATE_NEW);
+        this.linkReplacements.push(linkedContract);
+    }
+
+    /**
+     * Returns the contract's address as detemined by the provided AddressGenerator.
```
deepesh-kn: Typo: `detemined` -> `determined`

tools/deployment_tool/contract.js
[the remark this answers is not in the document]
deepesh-kn: I too prefer separate functions like `setAddress` and `generateAddress`

tools/deployment_tool/test/deployment_tools.test.js
```diff
@@ -0,0 +1,171 @@
...
+            const MerklePatriciaProof = Contract.loadTruffleContract('MerklePatriciaProof', null, { rootDir });
+            const MessageBus = Contract.loadTruffleContract('MessageBus', null, { rootDir });
+            MessageBus.addLinkedDependency(MerklePatriciaProof);
+
+            // provided ordering = correct ordering
+            const registry = new ContractRegistry();
+            registry.addContracts([MerklePatriciaProof, MessageBus]);
+
+            const output = registry.toParityGenesisAccounts();
+            assert.equal(Object.keys(output)[0], MerklePatriciaProof.getAddress());
```
deepesh-kn: We generally use strictEqual for the asserts.
Assert reason/messages are missing in all the assert statements.

hobofan:
@deepesh-kn @schemar

I addressed the feedback in the latest commit. `setAddress` has been split up into `setAddress` and `setGeneratedAddress`.

Issue #379, opened 2018-10-12
Gateway interacts with Mosaic Core


schemar:
The gateway contracts that we have should be updated to take the state root from the mosaic core contracts.

The interfaces of the gateway core and the core core should be unified into a single interface that is implemented by all cores and consumed by the gateways. That way, a decision how to get the state root is specific to each instance of the gateway, not its implementation.

schemar:
In this ticket we assume that Core has the state root in storage. But Core must implement the interface that Gateway consumes.

schemar:
Is #398 a duplicate of this?

PR #463, opened 2018-11-16
Gateway interacts with Mosaic Core

deepesh-kn:
fixes: #379.

This PR defines the StateRootInterface, which will be implemented by the gateway core and the core core.
In OriginCore two variables are declared `latestStateRootBlockHeight` and `stateRoots`.
`latestStateRootBlockHeight` store the block height of the latest commited stateRoot of auxiliary chian. `stateRoots` is the mapping that stores the stateroot of auxiliary chain against the block height.

This PR is only related to the defining the StateRootInterface and its implementation in OriginCore. The populating of `latestStateRootBlockHeight` and `stateRoots` are not part of this PR.

I have also implemented the test cases. These tests only verify that the data is received from the variables are correct. Reviewer can suggest if these tests are actually needed or not.


deepesh-kn:
The stateRoot stored are from the auxiliary chain for a particular block at height `h`. 
So there can be a stateRoot per commited meta-block. 
It is stored in the mapping against the block height of auxiliary chain blockheight. 
It is not origin chain block height or meta-block height.

Also `latestStateRootBlockHeight` stores the latest auxiliary block height, which can be queried by the faciliator/validators and they can use this block height in auxiliary chain to generate merkle proof.

Issue #450, opened 2018-11-06
Fix compiler warnings


schemar:
When compiling, there should be no warnings.

schemar:
If it's a method that is not yet implemented, you can add `_argumentName;` to the function body to keep the argument names and documentation.
In the case where the argument is actually unused, e.g. in a mock, you can also remove the name from the function argument and only leave the type.

PR #467, opened 2018-11-19
Fix compiler warnings

deepesh-kn:
Fixes: #450
This PR fixes all the compiler warnings.

Following were the types of warning.
1. Overshadowing of variable
2. Unused function parameter.
3. Uninitialized storage pointer.
4. Function state mutability can be restricted to pure
5. Functions in interfaces should be declared external.

Function `proveBlockOpening` was public, this had a compilation warning `Functions in interfaces should be declared external.` to fix this compilation warning I changed to external, this, in turn, resulted in to stack too deep error.
Even with commenting all the code, this error persisted. So I moved the proving origin core contract code to separate function and now it works. 

Another issue that was fixed was `Function state mutability can be restricted to pure` this was on `verify` function in KernelGateway contract which was pure. The TestKernelGateway contract overrides the function and could not become pure (check the implementation details). So to fix this now this function in TestKernelGateway just returns `true` and so it becomes pure. To test the fail case another contract `TestKernelGatewayFail` inherits `TestKernelGateway` and overrides the verify function to return `false`.


contracts/test/MockCore.sol
```diff
@@ -40,19 +40,19 @@ contract MockCore is Core {
...
 	  *
-	  * @return bytes32 storage root
+	  * @return bytes32 Mocked value of storage root.
 	  */
 	function getStorageRoot(
 		uint256 _blockHeight
 	)
 		public
-		view
+		pure
```
sarvesh-ost: Did you change this to fix compiler warning? 🤔  
If yes, you probably want to fix same for `getStateRoot` also, right?

test/core/kernel_gateway/prove_block_opening.js
```diff
@@ -38,53 +39,24 @@ contract('KernelGateway.proveBlockOpening()', async (accounts) => {
...
     auxiliaryBlockHash,
     storageBranchRlp,
-    accountRlp,
-    accountBranchRlp,
     originBlockHeight,
     kernelHash,
     transitionHash;
 
-  async function setAccountProof(isValid) {
-    accountRlp = testData.account.rlpEncodedAccount;
```
sarvesh-ost: `testData` is not more used. Remove from `require`.

contracts/core/KernelGateway.sol
```diff
@@ -166,6 +180,84 @@ contract KernelGateway is KernelGatewayInterface{
...
+     * @param _accountRlp RLP encoded account data.
+     * @param _accountBranchRlp RLP encoded path from root node to leaf node to
+     *                          prove account in merkle tree.
+     * @param _originBlockHeight The block height on origin where the opening
+     *                           took place. Must be a height that is finalised
+     *                           in the OriginBlockStore.
+     *
+     * @return success_ `true` if the proof is successful.
+     */
+    function proveOriginCore(
```
sarvesh-ost: This method is a potential duplicate of `proveGateway` method in `GatewayBase.sol`
[proveGateway](
https://github.com/OpenSTFoundation/mosaic-contracts/blob/617047f955b7a3bf88df7c5599650046ab282d99/contracts/gateway/GatewayBase.sol#L230)

I think logic to proven an account(even storage root `mapping`) should be extracted out in a contract. 

what do you think @schemar  @deepesh-kn  ?

contracts/core/KernelGateway.sol
```diff
@@ -37,13 +39,22 @@ contract KernelGateway is KernelGatewayInterface{
...
+    /** Emit event when proven open kernel is confirmed in auxiliary. */
     event OpenKernelConfirmed(
         bytes20 _originCoreIdentifier,
         bytes20 _auxiliaryCoreIdentifier,
         bytes32 _kernelHash,
         uint256 _currentDynasty
     );
 
+    /** Emit event origin core account is proven. */
+    event OriginCoreProven(
```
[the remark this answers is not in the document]
deepesh-kn: This is account proof of OriginCore contract. 
Storage root is one of the elements of account. So we are proving the existence of an account against a stateRoot.
We can say that indirectly that we are proving the storage root of an account. 
We can discuss this on call.

contracts/core/KernelGateway.sol
```diff
@@ -67,6 +78,9 @@ contract KernelGateway is KernelGatewayInterface{
     /** A mapping of kernel hash to the kernel object. */
     mapping(bytes32 => MetaBlock.Kernel) public kernels;
 
+    /** A mapping of origin block height to the storage root of origin core. */
+    mapping(uint256 => bytes32) public storageRoots;
```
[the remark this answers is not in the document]
deepesh-kn: Storage root is associated with an account. 
I am not sure if block store is right place.
I think KernelGateway is the right place as of now.

Also I was thinking about sarvesh's comment below. We may abstract the account proof in to separate contract and that manages the storage roots, just a thought.

contracts/core/KernelGateway.sol
```diff
@@ -166,6 +180,84 @@ contract KernelGateway is KernelGatewayInterface{
...
+                encodedOriginCorePath,
+                stateRoot
+            );
+
+            require(
+                storageRoot != bytes32(0),
+                "The storage root must not be zero."
+            );
+
+            storageRoots[_originBlockHeight] = storageRoot;
```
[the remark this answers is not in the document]
deepesh-kn: I have created an internal function `updateStorageRoot`.

contracts/core/KernelGateway.sol
```diff
@@ -166,6 +180,84 @@ contract KernelGateway is KernelGatewayInterface{
...
+            isAlreadyProven = true;
+
+        } else {
+
+            storageRoot = getStorageRoot(
+                _accountRlp,
+                _accountBranchRlp,
+                encodedOriginCorePath,
+                stateRoot
+            );
```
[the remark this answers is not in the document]
deepesh-kn: That's true. 
But its also proving origin core account, and hence storage root.
We can discuss this on call.

contracts/test/MockCore.sol
deepesh-kn: Compiler was not warning this statement. 
Apparantly while fixing `getStateRoot` I realised that when it is changed to `pure` it causes compilation error as its overridden function. So I just made it a proper view function. It used just dummy data.

Also I removed `getStorageRoot` it was not used anywhere.

test/core/kernel_gateway/prove_block_opening.js
deepesh-kn: Eagle eyes 👀

contracts/core/KernelGateway.sol
deepesh-kn: Good point. Let us create a ticket for this.

contracts/core/KernelGateway.sol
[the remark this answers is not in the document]
schemar: Agree!

contracts/core/KernelGateway.sol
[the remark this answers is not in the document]
schemar: Agree!

contracts/core/KernelGateway.sol
schemar: I also think we should create a ticket for this 👍

contracts/core/KernelGateway.sol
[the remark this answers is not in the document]
schemar: Agree!

test/core/kernel_gateway/prove_block_opening.js
schemar: 😳

contracts/core/KernelGateway.sol
deepesh-kn: Created a ticket for this.
https://github.com/OpenSTFoundation/mosaic-contracts/issues/478

PR #515, opened 2018-12-04
Gateways are now organized

### LINKED ISSUE #486: Gateway and CoGateway should be Organized.
- Gateway and CoGateway should be organised contracts.
- Remove the organisation address from the following contract constructors.
   - GatewayBase
   - Gateway
   - CoGateway 
   - EIP20Gateway
   - EIP20CoGateway
- Remove the `organisation` variable from GatewayBase.
- Write/Update the `constructors.js` tests for all the above contracts and also the modified functions.

schemar:
⚠️ First review and merge #514 ⚠️

I made the `GatewayBase` organized. As a result, the contracts that
directly or indirectly extend the `GatewayBase` must now provide an
address of a worker manager contract that implements the interface
`IsWorkerInterface`. All relevant modifiers were replaced by
`onlyWorker`.

Tests were updated to reflect the change. In order to easily test
organized contracts, I created a `MockWorkerManager` that can be used in
tests and has a single valid worker address.

I also re-organized the `contracts/test` directory a little bit with
additional sub-directories.

Fixes #486


contracts/gateway/EIP20CoGateway.sol
```diff
@@ -202,23 +203,23 @@ contract EIP20CoGateway is GatewayBase {
      * @param _utilityToken The utility token address that will be used for
      *                      minting the utility token.
      * @param _core Core contract address.
-     * @param _bounty The amount that facilitator will stake to initiate the
+     * @param _bounty The amount that facilitator will stakes to initiate the
```
deepesh-kn: 😳 
`The amount that facilitator stakes to initiate the stake process`

contracts/test/gateway/MockEIP20Gateway.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.5.0;
+
+import "../../gateway/EIP20Gateway.sol";
+import "../../lib/IsMemberInterface.sol";
+
+contract MockEIP20Gateway is EIP20Gateway {
```
deepesh-kn: I think after removing gateway linking this is not needed anymore.

contracts/gateway/EIP20CoGateway.sol
schemar: 👍

contracts/test/gateway/MockEIP20Gateway.sol
schemar: Wow good catch 🐙

PR #516, opened 2018-12-04
Fix Travis builds

schemar:
Trying to find the cause for the failing Travis builds


schemar:
Abandoned. Fixed in #520.

Issue #418, opened 2018-10-26
Restrict progressOutboxWithProof in MessageBus when outbox state of source chain is `RevocationDeclared` and inbox state of target chain is `Declared`.


deepesh-kn:
## User Story
If the message (Stake/Redeem) is *not* already progressed then Staker/Redeemer can revoke stake/redeem process.
There is no restriction as of now and anyone can progressOutboxWithProof and complete the process when the inbox message state is `Declared`. 
This is not the message initiator (Staker or Redeemer) intended and this will result into penalty loss.
Message initiator must have a complete control in such a scenario.

## Info
- While executing `progressOutboxWithProof` check the inbox message states. 
- If inbox message state is `Declared` and outbox message state is `RevocationDeclared` then `progressOutboxWithProof` should fail in this case.

We can also discuss if this can be allowed if the msg.sender is the message initiator.

PR #526, opened 2018-12-06
[WIP] Restrict progressOutboxWithProof in MessageBus when outbox state of source chain is `RevocationDeclared` and inbox state of target chain is `Declared`.

deepesh-kn:
⚠️Strange travis fail issue. May take some time, don't merge ⚠️

The actual issue change is trivial.
Real complexity here is the testing of the MessageBus.
MessageBus is an external library which further itself is lined with an external library `MerkelPatriciaProof`. This makes unit testing very difficult.

Currently to handle the unit tests the following approach is used.
- MockMerklePatriciaProof
This mocks `verify` function and always returns `true`.
- MockMerklePatriciaProofFail
This mocks `verify` function and always returns `false`.
- MockMessageBus
This library is an exact copy of `MessageBus` except that it is linked with `MockMerklePatriciaProof` library so that the Merkle proof verification can be mocked to return true.
- MockMessageBusFail
This library is an exact copy of `MessageBus` except that it is linked with `MockMerklePatriciaProofFail` library so that the Merkle proof verification can be mocked to return false.
- MessageBusWrapper
This is just an wrapper class to call the MessageBus library for testing. This contract is linked with `MockMessageBus`
- MessageBusWrapperFail
This is just an wrapper class to call the MessageBus library for testing. This contract is linked with `MockMessageBusFail`

Since the tests uses multiple copies of the file. We can run in to potential risk of Mock contracts/Libraries not being in sync with original files.

We have 2 options here.
- We just rely on integration test for Message bus.
- We write an addition test script that verifies that the Mock files are in sync with the originial file.
The syncing can be done in 2 ways. First by comparing of original and mocked files. Second genering mocked files on run time before running the unit tests.

If you have any suggestion to handle such tests in better way please let me know.
 

Fixes: #418


deepesh-kn:
This approach is not correct as we have 2 mock message bus and 2 MockMerklePatriciaProof library. Its decided to do an integration test for this. So closing this PR and moving the ticket to sprint backlog

Issue #417, opened 2018-10-26
Unit test all the functions of OSTPrime.


deepesh-kn:
## This story cover the following.
- Add unit tests for all the functions. (Please refer Utilitytoken test in openst-contract)
- Test of wrap and unwrap is already part of another ticket.
- Add Error message for requires if it is missing.

schemar:
Please make sure that you fully test `claim` and `redeem` (`wrap` and `unwrap`).

Specifically:
```solidity
// Line 138 should probably not be `transfer`, but rather:
transferFrom(msg.sender, address(this), _amount);

// Line 163 should probably not be `transferFrom`, but rather:
transfer(msg.sender, _amount);
```

Looks like a bug.

PR #533, opened 2018-12-10
Updated OSTPrime contract and tests.

deepesh-kn:
fixes: #417 
- This PR includes the unit test of all the functions of OST prime.
- The `Claim` event is renamed to `TokenUnwrapped`.
- The `Redeem` event is renamed to `TokenWrapped`.
-` uint256 _totalSupply` and  `address _utilityToken` params are removed from both the events as it is not needed.
-  Added a check so that`initialize` function can be called only once.
- `claim` function is renamed to `unwrap`.
- `redeem` function is renamed to `wrap`. 
- Both `wrap` and `unwrap` functions are modified.
- Unit tests are covered for all the functions of OST prime.
Please note: The unit tests for UtilityToken and EIP20Token is not covered in this. They are the separate tests.


contracts/test/gateway/TestOSTPrime.sol
```diff
@@ -0,0 +1,74 @@
...
+
+
+    /* Constructor */
+
+    /**
+     * @notice Contract constructor.
+     *
+     * @dev This contract is used only for testing.
+     *
+     * @param _valueToken ERC20 token address in origin chain
```
gulshanvasnani: Full-stop at the end of the sentence.

contracts/test/gateway/TestOSTPrime.sol
```diff
@@ -0,0 +1,74 @@
...
+
+import "../../gateway/OSTPrime.sol";
+
+/**
+ *  @title Test OST prime contract.
+ */
+contract TestOSTPrime is OSTPrime {
+
+
+    /* Constructor */
```
gulshanvasnani: Constructor ==> Special Functions

test/gateway/ost_prime/constructor.js
```diff
@@ -0,0 +1,81 @@
+const OSTPrime = artifacts.require("OSTPrime")
+  , BN = require('bn.js')
+  , web3 = require('../../test_lib/web3.js');
+
+const Utils = require('../../../test/test_lib/utils');
+
+const NullAddress = "0x0000000000000000000000000000000000000000";
```
gulshanvasnani: utils.js has "NullAddress" variable which can be used instead of defining a new variable here.

test/gateway/ost_prime/constructor.js
```diff
@@ -0,0 +1,81 @@
...
+contract('OSTPrime.sol', function (accounts) {
```
gulshanvasnani: We use contract name followed by the method being tested.
Example :-
Contract('OSTPrime.constructor()',function (accounts) {
});

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
+const OSTPrime = artifacts.require("OSTPrime")
+  , BN = require('bn.js');
+
+const Utils = require('../../../test/test_lib/utils');
+
+contract('OSTPrime.sol', function (accounts) {
```
gulshanvasnani: Same as last above.

test/gateway/ost_prime/constructor.js
```diff
@@ -0,0 +1,81 @@
...
+        symbol,
+        TOKEN_SYMBOL,
+        `Token symbol from contract must be ${TOKEN_SYMBOL}`
+      );
+
+      let decimals = await ostPrime.decimals.call();
+      assert.strictEqual(
+        TOKEN_DECIMALS.eq(decimals),
+        true,
+        `Token decimals from contract must be ${TOKEN_DECIMALS}`
```
gulshanvasnani: Comma at last of each line.
Example :-
assert.strictEqual(
        TOKEN_DECIMALS.eq(decimals),
        true,
        `Token decimals from contract must be ${TOKEN_DECIMALS}`,
);
Please follow it in all the files.

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    it('Should emit transfer event', async function () {
+      await initialize();
+
+      let tx = await ostPrime.unwrap(amount, { from: callerAddress });
+
+      let event = EventDecoder.getEvents(tx, ostPrime);
+
+      assert.strictEqual(
```
gulshanvasnani: Use assert.notStrictEqual instead of assert.strictEqual.

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    it('Should emit token unwrapped event', async function () {
+      await initialize();
+
+      let tx = await ostPrime.unwrap(amount, { from: callerAddress });
+
+      let event = EventDecoder.getEvents(tx, ostPrime);
+
+      assert.strictEqual(
```
gulshanvasnani: Same as above.

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+    });
+
+    it('Should emit transfer event', async function () {
+      await initialize();
+
+      let tx = await ostPrime.wrap({from: callerAddress, value: amount});
+
+      let event = EventDecoder.getEvents(tx, ostPrime);
+
+      assert.strictEqual(
```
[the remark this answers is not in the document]
deepesh-kn: Since it is for `undefined`. Updated code to use `assert.isDefined`

contracts/gateway/OSTPrime.sol
```diff
@@ -77,11 +73,15 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
+        require(
+            initialized == true,
+            "Contract is not initialized."
+        );
         _;
     }
 
-    /* Public functions */
+
+    /* Constructor */
```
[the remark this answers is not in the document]
deepesh-kn: We are following `/* Constructor */` in our code base. This is used for the constructor only.
Can you give me an example of special functions?

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
      *      On setup of the auxiliary chain the base tokens need to be
      *      transferred in full to OSTPrime for the base tokens to be
      *      minted as OST'
+     *
+     * @return success_ `true` if initialize was successful.
```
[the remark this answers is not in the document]
deepesh-kn: I am not sure. I think `true` will generate proper documentation.
@schemar, @sarvesh-ost  what is your opinion?

test/gateway/ost_prime/constructor.js
deepesh-kn: `NullAddress` is not exposed. If we all agree we can create a new ticket for this.

test/gateway/ost_prime/constructor.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    it('should pass with right set of parameters', async function () {
+
+      ostPrime = await OSTPrime.new(brandedTokenAddress);
+
+      assert.strictEqual(
+        web3.utils.isAddress(ostPrime.address),
+        true,
+        "OSTPrime contract address must not be zero"
```
schemar: Is this required? Would `ostPrime = await OSTPrime.new(brandedTokenAddress);` not fail?

test/gateway/ost_prime/constructor.js
```diff
@@ -0,0 +1,81 @@
...
+      let initialized = await ostPrime.initialized.call();
+      assert.strictEqual(
+        initialized,
+        false,
+        `Contract should not be initialized`
+      );
+
+    });
+
+    it('should fail if branded token address is not zero', async function () {
```
schemar:
```suggestion
    it('should fail if branded token address is zero', async function () {
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      brandedTokenAddress = accounts[2];
+      ostPrime = await OSTPrime.new(brandedTokenAddress);
+    });
+
+    it('The balance of OST prime contract must be zero', async function () {
+
+      let balance = await Utils.getBalance(ostPrime.address);
+      assert.strictEqual(
+        new BN(0).eq(new BN(balance)),
+        true,
```
schemar: 1. You don't need to convert to `BN` if none of the values to compare were `BN` before.
2. *If* you use `BN`, you can `assert(one.eq(two), 'message')`

It seems this should simply be:
```js
      let balance = await Utils.getBalance(ostPrime.address);
      assert.strictEqual(
        0,
        balance,
        `The balance of the contract must be zero after initialization.`
      );
```

Generally, I am not sure this test is sensible 🤔

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+        `Contract should be initialized.`
+      );
+
+    });
+
+    it('should fail if already initialized', async function () {
+
+      await ostPrime.initialize(
+        {from: accounts[2], value:TOKENS_MAX}
```
schemar:
```suggestion
        {from: accounts[2], value: TOKENS_MAX}
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      );
+
+      await Utils.expectRevert(
+        ostPrime.initialize.call(
+          {from: accounts[3], value:TOKENS_MAX}
```
schemar:
```suggestion
          {from: accounts[3], value: TOKENS_MAX}
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('should fail if already initialized', async function () {
+
+      await ostPrime.initialize(
+        {from: accounts[2], value:TOKENS_MAX}
+      );
+
+      await Utils.expectRevert(
+        ostPrime.initialize.call(
```
schemar: Why `.call` here?
```suggestion
        ostPrime.initialize(
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('should fail if the initialize is called with payable amount not equal ' +
+      'to 800M ost prime base token', async function () {
+
+      let tokenAmount = new BN(1);
+
+      await Utils.expectRevert(
+        ostPrime.initialize.call(
+          {from: accounts[2], value:tokenAmount}
```
schemar:
```suggestion
          {from: accounts[2], value: tokenAmount}
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      assert.strictEqual(
+        new BN(0).eq(new BN(balance)),
+        true,
+        `The balance of contract must be zero.`
+      );
+
+    });
+
+    it('should fail if the initialize is called with payable amount not equal ' +
+      'to 800M ost prime base token', async function () {
```
schemar:
```suggestion
      'to TOKENS_MAX ost prime base token', async function () {
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      await Utils.expectRevert(
+        ostPrime.initialize.call(
+          {from: accounts[2], value:tokenAmount}
+        ),
+        "Payable amount must be equal to total supply of token."
+      );
+
+    });
+
+    it('should pass if the initialize is called with value equal than 800M ' +
```
schemar:
```suggestion
    it('should pass if initialize is called with value equal to TOKENS_MAX ' +
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      'ost prime base token', async function () {
+      let result = await ostPrime.initialize.call(
+        {from: accounts[2], value:TOKENS_MAX}
```
schemar:
```suggestion
        {from: accounts[2], value: TOKENS_MAX}
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      );
+
+      assert.strictEqual(
+        result,
+        true,
+        `The result should be true.`
+      );
+
+      await ostPrime.initialize(
+        {from: accounts[2], value:TOKENS_MAX}
```
schemar:
```suggestion
        {from: accounts[2], value: TOKENS_MAX}
```

test/gateway/ost_prime/initialize.js
```diff
@@ -0,0 +1,96 @@
...
+      );
+
+      let balance = await Utils.getBalance(ostPrime.address);
+      assert.strictEqual(
+        TOKENS_MAX.eq(new BN(balance)),
```
schemar: `BN::eqn` to compare to numbers (non-BN).
```suggestion
        TOKENS_MAX.eqn(balance),
```

Again, `assert()` instead of `assert.strictEqual()`?

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
+     * @notice convert the OST' EIP-20 token to OST` base token.
      *
-     * @param _amount Amount of basetoken
+     * @param _amount Amount of EIP-20 to convert to base token.
      *
-     * @return `true` if claim was successfully progressed
+     * @return success_ `true` if unwrap was successful.
      */
-    function claim(
+    function unwrap(
```
schemar: `wrap` and `unwrap` should probably be `external`

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+      await ostPrime.setTokenBalance(callerAddress, amount);
+
+    });
+
+    it('Should fail when the amount is zero', async function () {
+      await initialize();
+
+      let amount = 0;
+      await Utils.expectRevert(
+        ostPrime.unwrap.call(amount, { from: callerAddress }),
```
schemar: Why `.call`? (also other test cases where a `revert` is expected)

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+        "Amount should not be zero."
```
schemar: I think we use single quotes in JS (all files).

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+      ostPrime = await OSTPrime.new(brandedTokenAddress);
+
+      callerAddress = accounts[3];
+      amount = new BN(500);
+
+      await ostPrime.setTokenBalance(callerAddress, amount);
+
+    });
+
+    it('Should fail when the amount is zero', async function () {
```
schemar: We start all `it()`s with lowercase (see also other function calls)
```suggestion
    it('should fail when the amount is zero', async function () {
```

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+      await initialize();
+
+      let amount = new BN(501);
+      await Utils.expectRevert(
+        ostPrime.unwrap.call(amount, { from: callerAddress }),
+        "Insufficient balance."
+      );
+
+    });
+
```
schemar: It seems you are missing `it('should fail when the amount is greater than the contract's base token balance')`

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+      );
+
+      let tx = await ostPrime.unwrap(amount, { from: callerAddress });
+      let gasUsed = new BN(tx.receipt.gasUsed);
+
+      let callerEIP20Tokenbalance = await ostPrime.balanceOf.call(callerAddress);
+      assert.strictEqual(
+        callerEIP20Tokenbalance.eq(new BN(0)),
+        true,
+        `The balance of ${callerAddress} should be zero.`
```
schemar: Same as other balance comparisons

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+
+      assert.strictEqual(
+        finalContractBalance.eq(initialContractBalance.sub(amount)),
+        true,
+        `Contract base token balance should decrease by ${amount}`
+      );
+
+      assert.strictEqual(
+        finalCallerBalance.eq(initialCallerBalance.add(amount).sub(gasUsed)),
+        true,
```
schemar: More `BN` asserts throughout all test files

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,192 @@
...
+        `The contract should return true.`
+      );
+
+      let tx = await ostPrime.unwrap(amount, { from: callerAddress });
+      let gasUsed = new BN(tx.receipt.gasUsed);
+
+      let callerEIP20Tokenbalance = await ostPrime.balanceOf.call(callerAddress);
+      assert.strictEqual(
+        callerEIP20Tokenbalance.eq(new BN(0)),
```
schemar: I would test an amount that leaves another token balance `>0` and assert that that is correct. Seems like a better test to me that way.

test/gateway/ost_prime/unwrap.js
schemar: or `isDefined()` https://www.chaijs.com/api/assert/

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
+const OSTPrime = artifacts.require("TestOSTPrime")
+  , BN = require('bn.js');
+
+const Utils = require('../../../test/test_lib/utils');
+const EventDecoder = require('../../test_lib/event_decoder.js');
+
+contract('OSTPrime.sol', function (accounts) {
+
+  describe('Unwrap', async () => {
```
schemar: `wrap` and `contract`

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+    const DECIMAL = new BN(10);
+    const POW = new BN(18);
+    const DECIMAL_FACTOR = DECIMAL.pow(POW);
+    const TOKENS_MAX = new BN(800000000).mul(DECIMAL_FACTOR);
+
+    let brandedTokenAddress, ostPrime, callerAddress, amount;
+
+    async function initialize(){
+      await ostPrime.initialize(
+        {from: accounts[5], value:TOKENS_MAX}
```
schemar:
```suggestion
        {from: accounts[5], value: TOKENS_MAX}
```

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+      );
+    };
+
+    beforeEach(async function () {
+
+      brandedTokenAddress = accounts[2];
+      ostPrime = await OSTPrime.new(brandedTokenAddress);
+
+      callerAddress = accounts[3];
+      amount = new BN(500);
```
schemar: It would be relevant for this test what the balance of `accounts[3]` is. Unfortunately, nothing can be assumed as other tests might have changed the balance before this test runs. It is not well decoupled that way.

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+        "Available balance is less than payable amount."
+      );
+
+    });
+
+    it('Should fail when EIP-20 balance is insufficient', async function () {
+
+      await initialize();
+
+      // @dev this test with .call does not revert as expected.
```
schemar: But you still use `call()`? (why)

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+      );
+
+    });
+
+    it('Should fail when the payable amount sent is less than the available ' +
+      'balance', async function () {
+
+      await initialize();
+
+      // @dev this test with .call does not revert as expected.
```
schemar: Because `.call` does not send any value and the balance is still available on the account, right? Also, at this point in time the balance of `accounts[9]` could be anything and should be used to deduct which amount should be used.

Also, like mentioned in the `OSTPrime` contract, I think a transfer of more than 50% of an account's balance would fail even though it should pass.

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+        );
+
+      assert.strictEqual(
+        result,
+        true,
+        `The contract should return true.`
+      );
+
+      let tx = await ostPrime.wrap({from: callerAddress, value: amount});
+      let gasUsed = new BN(tx.receipt.gasUsed);
```
schemar: It seems balances/gas amounts don't need to be wrapped in `BN`s generally as they cannot be that large. Not sure. But we can always use the BN API for comparisons (`eq`, `eqn`, `eqrn`, ...) https://github.com/indutny/bn.js/#notation

test/gateway/ost_prime/wrap.js
```diff
@@ -0,0 +1,209 @@
...
+
+      assert.strictEqual(
+        amount.eq(eventData._value),
+        true,
+        `The _value in the event should be equal to ${amount}`
+      );
+
+    });
+
+    it('Should emit token unwrapped event', async function () {
```
schemar:
```suggestion
    it('Should emit token wrapped event', async function () {
```

tools/runGanacheCli.sh
```diff
@@ -9,7 +9,7 @@ node_modules/.bin/ganache-cli \
     --account="0x8fbbbaceff30d4eea3e2ffa2dfedc3c053f78c1f53103e4ddc31309e6b1d5ea6,10000000000000000000000000000000" \
     --account="0x8fbbbaceff30d4eea3e2ffa2dfedc3c053f78c1f53103e4ddc31309e6b1d5ea7,10000000000000000000000000000000" \
     --account="0x8fbbbaceff30d4eea3e2ffa2dfedc3c053f78c1f53103e4ddc31309e6b1d5ea8,10000000000000000000000000000000" \
-    --account="0x8fbbbaceff30d4eea3e2ffa2dfedc3c053f78c1f53103e4ddc31309e6b1d5ea9,0" \
+    --account="0x8fbbbaceff30d4eea3e2ffa2dfedc3c053f78c1f53103e4ddc31309e6b1d5ea9,12000500" \
```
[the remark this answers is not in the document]
schemar: A agree, this change potentially messes things up :(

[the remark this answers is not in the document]
deepesh-kn: Done

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
+            "Amount should not be zero."
+        );
+
         require(
             _amount <= balances[msg.sender],
-            "Insufficient balance"
+            "Insufficient balance."
         );
 
         assert(address(this).balance >= _amount);
```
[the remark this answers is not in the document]
deepesh-kn: This is `assert`,  this condition should never happen in production.
Also `assert` does not support the reason string.

I have converted this to require. I personally think that this should be an assert.

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
 
-        transfer(address(this),_amount);
+
+        balances[msg.sender] = balances[msg.sender].sub(_amount);
+        balances[address(this)] = balances[address(this)].add(_amount);
```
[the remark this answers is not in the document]
deepesh-kn: We can discuss this. Using transferFrom is messy.
- Currently transfer is public function, we cannot call it as msg.sender will change.
- We need to create an internal function with transfer code, so that it can be called by this class and also from the public tranferFrom function.
- TransferFrom requires the approval. This will anyhow make the code duplicate.

I tried it in my first approach but I felt that this looks better. Although it has 3 lines of duplicate code. Let me know if you disagree, I will make this change again.

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
+        emit Transfer(msg.sender, address(this), _amount);
+        emit TokenUnwrapped(msg.sender, _amount);
 
-        return true;
+        success_ = true;
     }
 
     /**
-     * @notice convert the base token to OST utility token
+     * @notice convert OST` base token to OST EIP-20 token.
```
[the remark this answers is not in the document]
deepesh-kn: Removed EIP-20 from all the comments.

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
         public
         onlyInitialized
         payable
-        returns (bool /** success */)
+        returns (bool success_)
     {
-        require(msg.value == _amount);
-        assert(address(this).balance >= _amount);
+        require(
+            msg.value > 0,
```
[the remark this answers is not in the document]
deepesh-kn: Done.
Also for msg.sender.

test/gateway/ost_prime/constructor.js
deepesh-kn: Even if `ostPrime = await OSTPrime.new(brandedTokenAddress);` fails, tests will continue as there is no asserts.
This is helpful to assert that deployment is failed rather than getting the errors from the other test failure.

test/gateway/ost_prime/initialize.js
deepesh-kn: Initialize method requires that the payable amount must be equal to the TOKEN_MAX. 
So shouldn't it be tested that the initial amount before the initialization should be zero? Else after initialization, the contract may contain more than TOKEN_MAX.
Let me know if this sounds sensible, If not I can remove this.

About converting to BN:

That will not work. `getBalance` returns `string`. 
Changed to 
```
assert.strictEqual(
      balance,
      '0',
      `The balance of contract must be zero.`,
    );
```

test/gateway/ost_prime/initialize.js
deepesh-kn: balance is `string`, so its converted to `BN`. 
This is already using `assert.strictEqual`.
```
let balance = new BN(await Utils.getBalance(ostPrime.address));
    assert.strictEqual(
      TOKENS_MAX.eq(balance),
      true,
      `The balance of contract must be ${TOKENS_MAX}.`,
    );
```

test/gateway/ost_prime/unwrap.js
deepesh-kn: `initialCallerBalance` is going to be a big number and I need to perform arithmetic operation. 🤔

test/gateway/ost_prime/wrap.js
deepesh-kn: True!
Accounts have 10000000000000000000000000000000 balance.
Which is still enough considering 800M ether as max gas to run the tests (Considering other tests are just consuming gas and not moving the balance in large amount).

Also the tests are considering the initial balances while asserting the final balances.

test/gateway/ost_prime/wrap.js
deepesh-kn: True, but gasUsed is used as follows, so its converted to BN.
```assert.strictEqual(
      finalCallerBalance.eq(initialCallerBalance.sub(amount).sub(gasUsed)),
      true,
      `Caller's base token balance should decrease by ${amount.sub(gasUsed)}`,
    );
```

contracts/gateway/OSTPrime.sol
```diff
@@ -104,65 +110,109 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
 
-        allowed[address(this)][msg.sender] = _amount;
-        transferFrom(address(this), msg.sender, _amount);
+        require(
+            amount > 0,
+            "Payable amount should not be zero."
+        );
+
+        require(
+            address(account).balance >= amount,
```
sarvesh-ost: I didn't understand this part. 
Theoretically, account eth balance should be reduced when this transaction is triggered. By this logic, the caller can't wrap all the amount. 

A unit test case to wrap the total balance amount should validate this. 

@schemar  @gulshanvasnani  What do you think?

contracts/gateway/OSTPrime.sol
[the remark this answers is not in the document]
deepesh-kn: We continue to use the `constructor` for now. This we can take up when we have a unified style guide for OST.

contracts/gateway/OSTPrime.sol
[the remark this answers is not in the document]
deepesh-kn: Reverted to assert. Added a comment.

contracts/gateway/OSTPrime.sol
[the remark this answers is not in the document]
deepesh-kn: Done. Implemented `transferBalance` function.

contracts/gateway/OSTPrime.sol
```diff
@@ -105,64 +112,99 @@ contract OSTPrime is UtilityToken, OSTPrimeConfig {
...
-        assert(address(this).balance >= _amount);
+        require(
+            msg.value > 0,
+            "Payable amount should not be zero."
+        );
 
-        allowed[address(this)][msg.sender] = _amount;
-        transferFrom(address(this), msg.sender, _amount);
+        require(
+            address(msg.sender).balance >= msg.value,
```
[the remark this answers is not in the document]
deepesh-kn: Yes, you are correct, tested this again. This require is not needed.

test/gateway/ost_prime/constructor.js
deepesh-kn: You are correct, it fails 👍

test/gateway/ost_prime/unwrap.js
```diff
@@ -0,0 +1,222 @@
...
+      );
+
+    });
+
+  it('should fail when the amount is greater than the contract\'s base ' +
+    'token balance', async function () {
+
+    await initialize();
+
+    let amount = new BN(1).add(TOKENS_MAX);
```
schemar: I think `TOKENS_MAX.addn(1)` would read easier.

test/test_lib/utils.js
```diff
@@ -166,16 +166,24 @@ Utils.prototype = {
...
     try {
       await promise;
     } catch (error) {
       assert(
-        error.message.search('invalid opcode') > -1,
+        error.message.search('Returned error:') > -1,
         'The contract should fail an assert. Instead: ' + error.message
       );
 
+      assertExpectedMessage(expectedMessage, error);
```
schemar: What kind of message would you expect with an assert? An assert would always return `invalid opcode`, or not?

deepesh-kn: `invalid opcode` is one of the assertion failures.
We are also interested in the `sender doesn't have enough funds to send tx.` error

deepesh-kn:
> @deepesh-kn what is your confidence level now?

My confidence level is very good.
What do you think as a reviewer?

gulshanvasnani:
LGTM
Great Work :thumbsup:

test/test_lib/utils.js
schemar: 👍

PR #541, opened 2018-12-12
Updated EIP20Gateway constructor parameters

gulshanvasnani:
This PR contains following :-
1.  Verifying the membersmanager address.
2. Added test case which will fail when address of membersmanager is passed as zero.
Fixes:- #422


test/gateway/eip20_gateway/constructor.js
```diff
@@ -150,6 +157,22 @@ contract('EIP20Gateway.constructor() ', function (accounts) {
         );
 
     });
+    
+    it('should fail when members manager address is passed as zero', async function () {
```
schemar: This looks like it belongs to the `Organized` tests and not the `EIP20Gateway` tests. If we change `Organized` and make the members manager optional, now we have to update test cases for the EIP-20 Gateway, which we didn't actually touch.

sarvesh-ost: I think it's fine to keep this test. 
Because there is no other way to test that child contract calls the constructor of the parent.

deepesh-kn: For constructors and overridden functions, I think we should have the test.

schemar: OK if you want, but it is not correct to have this test here.

PR #542, opened 2018-12-12
Accept burner address in the gateway and cogateway

### LINKED ISSUE #525: Gateways should have a "burner" address
Currently, burning in the gateways (`Gateway` and `CoGateway`) is hardcoded as transfers to `address(0)`.
Instead, this address should be configurable as the "burner" address on construction and not be changeable afterwards, also not by organization.
Whenever a gateway is supposed to burn tokens, it should send it to the set "burner" address.

This means `OSTPrime` does not require a `burn` function, only the known `increaseSupply` and `decreaseSupply` functions.

For Safe Mosaic we will always set this to the `0x0` address. Later, burner contracts will handle the burning; see #532 for details.

sarvesh-ost:
Fixes #525  

Gateway and Co-gateway now accept burner address in the constructor. Burner address is used to burn penalty and bounty in case of `progressRevertStake` and `progressRevertRedeem`

### DIFF
--- a/contracts/gateway/EIP20CoGateway.sol
+++ b/contracts/gateway/EIP20CoGateway.sol
@@ -186,6 +186,9 @@ contract EIP20CoGateway is GatewayBase {
     /** address of value token. */
     address public valueToken;
 
+    /** Address where token will be burned. */
+    address payable public burner;
+
     /** Maps messageHash to the Mint object. */
     mapping(bytes32 /*messageHash*/ => Mint) mints;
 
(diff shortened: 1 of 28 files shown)


sarvesh-ost:
@schemar  Those tests will be covered in below stories and this was also initially agreed during planning. 
Refer:  #426 and #440

schemar:
> @schemar Those tests will be covered in below stories and this was also initially agreed during planning.
> Refer: #426 and #440

OK if you think it makes sense to split writing tests from writing the implementation.

Issue #543, opened 2018-12-12
Organization should be more general


schemar:
The organization and its related contracts should be identical in the `openst` and the `mosaic` repository. For that we need a generalized organization.

## Todos

- [x] Organized modifiers
- [x] Restructure organization interface
- [x] Make admin internal and give it all rights of owner
- [x] Delete `OpsManaged` and `Workers`
- [x] Replace `isOwner` with `isOrganization`
- [x] Remove assignment restrictions (documentation!)
- [x] Remove `Owned`
- [x] Initialize on construction

## Organization

The organization should no longer leak the internals of an admin. From the outside (interface), it should only have the notion of "the organization" and "workers of the organization".
The admin in the organization is an implementation detail of this implementation of the organization where the owner can delegate all rights to an appointed admin.
This means that the `isOwner` method should instead be called `isOrganization`. It should return true for both cases where `msg.sender` is the owner and where `msg.sender` is the admin. The owner grants the admin all rights on the owner's behalf.

### Restrictions

Remove any existing restrictions that restrict a role to be assigned or not assigned to specific keys. For example, it should not be forbidden for the owner to be the same key as the admin.
However, in the documentation it should be made **very** clear that that is not a good idea in practice and that it is always a good idea to use different keys for owner, admin, and workers. 

### Interfaces

Remove the `OrganizationInterface`, which is not used anywhere, and instead make the `IsMemberInterface` the `OrganizationInterface` with the methods `isWorker` and `isOrganization`.

### Initialization

When constructing the organization contract, it should be possible to initialize it with a given owner, a given admin, an array of workers, and their expiration height. Only a single expiration height for all workers.

## Organized

Organized should have the modifiers `onlyWorker` and `onlyOrganization`. The organization implementation should not leak the fact that it can set an admin to act on behalf of the organization.

## ℹ️ Info
Once the organization is updated, a PR should be opened on [openst-contracts](https://github.com/OpenSTFoundation/openst-contracts) to apply the changes there as well

schemar:
This ticket is up for discussion and there are open questions. E.g.

* Do we need more modifiers (e.g. `onlyOwnerOrAdmin`)?
* Do the interface changes make sense?

Discussion will happen next week on Monday.

/cc @abhayks1 @jasonklein @pgev

PR #547, opened 2018-12-13
The organization now has a unified interface

schemar:
There is only an interface that represents an organization. That
interface exposes the methods expected to be found on an organization.
Tests were updated accordingly.

References #543


schemar:
Added @pgev as reviewer so we can be faster when moving the updates to `openst-cotracts`.

contracts/lib/Organization.sol
```diff
@@ -83,16 +117,40 @@ contract Organization is OrganizationInterface, IsMemberInterface {
     /* Constructor */
 
     /**
-     * @notice The address that creates the contract is set as the initial
-     *         owner.
+     * @notice
```
pgev: We should either remove empty notice, or add documentation.

contracts/lib/Organization.sol
```diff
@@ -83,16 +117,40 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
+            setWorkerInternal(_workers[i], _expirationHeight);
+        }
     }
 
 
-    /* External Functions */
+    /* External functions */
 
     /**
      * @notice Proposes a new owner of this contract. Ownership will not be
```
pgev: Let's remove return (success/failure) statement of the function.
Also, in other functions of the contract, we should remove the return statement if it is only success/failure.

contracts/lib/Organization.sol
```diff
@@ -83,16 +117,40 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
+        admin = _admin;
+
+        for(uint256 i = 0; i < _workers.length; i++) {
+            setWorkerInternal(_workers[i], _expirationHeight);
+        }
     }
 
 
-    /* External Functions */
+    /* External functions */
```
pgev: Functions ;)

contracts/lib/Organization.sol
```diff
@@ -197,20 +254,7 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
-        );
-
-        workers[_worker] = _expirationHeight;
-        remainingBlocks_ = _expirationHeight.sub(block.number);
-
-        emit WorkerSet(_worker, _expirationHeight, remainingBlocks_);
+        remainingBlocks_ = setWorkerInternal(_worker, _expirationHeight);
     }
 
     /**
```
pgev: I would suggest to remove isUnset variable at all. This costs gas and can be checked by caller if needed.

contracts/lib/Organization.sol
```diff
@@ -235,15 +279,25 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
-        isOwner_ = _owner == owner;
+    function isOrganization(
+        address _organization
+    )
+        external
+        view
+        returns (bool isOrganization_)
+    {
+        isOrganization_ = _organization == owner || _organization == admin;
     }
```
pgev: A proposal to rename the function to isActiveWorker as that's what we are checking, however, we could also assume that isWorker is a better interface for isActiveWorker. Thoughts?

contracts/lib/Organization.sol
```diff
@@ -260,4 +314,39 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
+            "Worker address cannot be null."
+        );
+
+        require(
+            _expirationHeight > block.number,
+            "Expiration height must be in the future."
+        );
+
+        workers[_worker] = _expirationHeight;
+        remainingBlocks_ = _expirationHeight.sub(block.number);
```
pgev: Let's not have this calculation in this function.

contracts/lib/Organization.sol
```diff
@@ -260,4 +314,39 @@ contract Organization is OrganizationInterface, IsMemberInterface {
...
+    /**
+     * @notice Sets worker and its expiration block height.
+     *
+     * @param _worker Worker address to be added.
+     * @param _expirationHeight Expiration block height of worker.
+     *
+     * @return remainingBlocks_ Remaining number of blocks for which worker is
+     *                          active.
+     */
+    function setWorkerInternal(
```
pgev: What about combining setWorker and setWorker internal functions as a one and make it public?

contracts/lib/Organization.sol
```diff
@@ -260,4 +314,39 @@ contract Organization is OrganizationInterface, IsMemberInterface {
         isWorker_ = workers[_worker] > block.number;
     }
 
+
+    /* Private functions */
+
+    /**
+     * @notice Sets worker and its expiration block height.
```
pgev: Let's also mention, that if worker exists, the function updates it's data (expiration height in this case).

contracts/lib/OrganizationInterface.sol
```diff
@@ -14,84 +14,36 @@ pragma solidity ^0.5.0;
...
-    function completeOwnershipTransfer() external returns (bool success_);
-
-    /**
-     * @notice Sets admin address.
+     * @param _worker Address to check.
      *
-     * @param _admin Admin address to be set.
-     *
-     * @return success_ True on successful execution.
+     * @return isWorker_ True if the worker is already added and expiration
```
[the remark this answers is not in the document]
schemar: ✅ definitely. Updated.

contracts/lib/Organization.sol
schemar: We actually have more `External functions` in the code base than `External Functions` ;)

contracts/lib/Organization.sol
schemar: Why? It is currently used in various places throughout the [repo](https://github.com/OpenSTFoundation/mosaic-contracts/search?q=success_) and it is even part of the [style guide](https://github.com/OpenSTFoundation/openst-contracts/blob/2b16c4191455aea6ed974b0d0f543e99870f1277/SOLIDITY_STYLE_GUIDE.md#alignment).
I remember @deepesh-kn or @sarvesh-ost had an argument for it 🧐
Feedback please @deepesh-kn / @sarvesh-ost

contracts/lib/Organization.sol
schemar: If you refer to `unsetWorker`: ✅ done
Otherwise please give context I am not sure this comment belongs in this line.

contracts/lib/Organization.sol
schemar: I think `isWorker` is a better interface. It is an implementation detail of this organization that workers can expire. Please resolve the comment if you agree.

contracts/lib/Organization.sol
schemar: Why?

contracts/lib/Organization.sol
schemar: 👍 done

contracts/lib/Organization.sol
schemar: How would that work with the `external` interface?

contracts/lib/Organization.sol
```diff
@@ -16,22 +16,51 @@ pragma solidity ^0.5.0;
...
+
+    /** Emitted when a worker address was set. */
+    event WorkerSet(
+        address indexed worker,
+        uint256 expirationHeight,
+        uint256 remainingHeight
+    );
+
+    /** Emitted when a worker address is deleted from the contract. */
+    event WorkerUnset(address worker, bool wasSet);
```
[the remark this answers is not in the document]
abhayks1: Workers can be set with an expirationHeight 0. To differentiate whether "0 expiration height or non existent worker" is being deleted or " valid worker with non 0 expiration height" is being deleted, wasSet flag was added.

contracts/lib/Organization.sol
```diff
@@ -16,22 +16,51 @@ pragma solidity ^0.5.0;
 
 import "./SafeMath.sol";
 import "./OrganizationInterface.sol";
-import "./IsMemberInterface.sol";
 
 
 /**
  * @title Organization contract.
```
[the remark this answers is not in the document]
pgev: If it's not a very obvious case than having a documentation is always useful. However, if it's not adding anything, then better to skip. Here, it's better to have.

contracts/lib/Organization.sol
```diff
@@ -16,22 +16,51 @@ pragma solidity ^0.5.0;
...
+    /* Events */
+
+    /** Emitted when a current owner initiates a change of ownership. */
+    event OwnershipTransferInitiated(address indexed proposedOwner);
+
+    /** Emitted when a new owner accepts the ownership transfer. */
+    event OwnershipTransferCompleted(address newOwner);
+
+    /** Emitted whenever an owner or admin changes the address of the admin. */
+    event AdminAddressChanged(address indexed newAdmin);
```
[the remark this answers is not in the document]
pgev: I think Admin*Address*Changed is inconsistent, as I did not see usage of *Address* in other cases. However, it's a very minor point and I am good to keep it us it is.

contracts/lib/Organization.sol
```diff
@@ -16,22 +16,51 @@ pragma solidity ^0.5.0;
...
+
+    /** Emitted when a worker address was set. */
+    event WorkerSet(
+        address indexed worker,
+        uint256 expirationHeight,
+        uint256 remainingHeight
```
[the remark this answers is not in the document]
pgev: Sorry for confusion. I would propose to *remove* remainingHeight parameter. Do not see a reason to have it.

contracts/lib/Organization.sol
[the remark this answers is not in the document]
pgev: If caller needs to differentiate those cases (*if*), it can check it before calling removing worker. Doing it for all costs gas and I believe is not needed for most of our cases.

contracts/lib/Organization.sol
pgev: Let's discuss this point (/cc @deepesh-kn @sarvesh-ost @schemar)

I think, using both error handling mechanisms (exception & error code) should not be common case. In solidity world, as there is no way to differentiate what type of exception was thrown (no catching) occasionally using return error code concept can be useful to handle function based on return value. However, this is rare case, and we could mostly rely on exception to be thrown in case of error.

contracts/lib/Organization.sol
pgev: My bad, I was referring to remainingBlocks parameter.

contracts/lib/Organization.sol
pgev: If `isWorker` would be part of interface, then yes, being active or expired is an implementation detail. However, expiration height is part of this contract interface (being part of its events and public functions), hence it's not an implementation detail. I am more inclined for renaming it to isActiveWorker, however, if you think isWorker is simpler and from documentation it's clear that it's about being registered and not expired, let's keep it.

contracts/lib/Organization.sol
pgev: It costs gas, and if consumer of the function would need this information, it can check the expiration height of the worker (call to workers storage variable) and calculate it for that concrete case.

contracts/lib/Organization.sol
pgev: That's why I suggest to make it `public`. It is an additional call we do each time.
However, I am open to listen why do you prefer it in this way.

contracts/lib/Organization.sol
schemar: ok. that's tracked in https://github.com/OpenSTFoundation/mosaic-contracts/pull/547/files#r242095075

contracts/lib/Organization.sol
schemar: Can it be declared `external` in the  interface and be implemented `public` in the contract?

contracts/lib/Organization.sol
schemar: tracked in https://github.com/OpenSTFoundation/mosaic-contracts/pull/547/files#r242095075

contracts/lib/Organization.sol
pgev: I am going to say yes (needs to be checked), as it's widening the accessibility scope of the function. Public methods can be called internally and externally, so we do not break the interface design of the interface contract.

contracts/lib/Organization.sol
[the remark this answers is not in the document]
abhayks1: Worker can be set with higher expiration height. remainingHeight ensures event listener knows from the current block, for how many blocks worker is active for. 
I know It can be derived from expirationHeight, however we concluded remainingHeight to be present in the event .
If we think anyone of  expirationHeight or remainingHeight should be there, we can remove expirationHeight and keep remainingHeight because expirationHeight is part of setWorker method signature.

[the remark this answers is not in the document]
schemar: By that argument we need neither. Expiration height is given and remaining height can be calculated. However, the events are not for the people who did the transactions, but for others to be notified about changes. In this case, I could imagine a worker wants to know that its key is now set as a valid worker. In that case, the worker would also be interested in the expiration height.
The remaining height (which we don't currently see as relevant) is simply `expirationHeight - blockNumber`, both of which are listed in the web3 log and are therefore easily accessible by anyone who reads the event.
Calculating and adding the "remaining height" parameter to the event costs gas that we can potentially save, unless someone can explain what it is for.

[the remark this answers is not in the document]
abhayks1: Discussed this point with Martin on quick call. Also tried to find in the history why remainingHeight was added with Workers contract. As far as I remember it was requested by the KIT at that time with Workers/Pricer/Airdrop integration. Looking at the current situation(gas cost) we can keep expirationHeight and remove remainingHeight as it can be derived from expirationHeight. We can always add it if it's needed in future with a separate ticket. 👍 🚗

contracts/lib/Organization.sol
sarvesh-ost: Keeping success/failure helps in unit testing. `Method.call` in truffle tests return `true` or `false` which is more natural assert than expecting no exception. I would prefer keeping this as it is. This is being followed in the codebase. 
Also, I think this discussion is beyond the scope of this PR.  can this be discussed separately? 🤔If there is any action item, we can always create a separate ticket.  😃

contracts/lib/Organization.sol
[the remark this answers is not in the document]
schemar: ✅ removed

contracts/lib/Organization.sol
schemar: It would indeed work! However, there is another problem: the external function should be restricted to `onlyOwnerOrAdmin` while the internal call should not (it is called on construction). Ideas?

contracts/lib/Organization.sol
deepesh-kn: I always found it useful to be deterministic about what to expect from the function.
@pgev, we should discuss this on call to understand the advantage and disadvantage for this. For this PR we should not do changes in all the files,  we should create a new ticket for this after discussion.

schemar: The only problem is that is has to be the same in `mosaic` and `openst`, so we have to make a decision for this function.

test/gateway/eip20_cogateway/confirm_stake_intent.js
```diff
@@ -356,7 +355,7 @@ contract('EIP20CoGateway.confirmStakeIntent() ', function (accounts) {
         "Merkle proof verification failed.",
       );
 
-  });
+    });
```
sarvesh-ost: The file doesn't have consistent spacing, either 2 or 4 not mix.

test/gateway/utility_token/set_co_gateway.js
```diff
@@ -75,20 +75,20 @@ contract('UtilityToken.setCoGateway() ', function (accounts) {
...
-      utilityToken.setCoGateway(coGateway.address, { from: owner }),
-      'CoGateway should be linked with this utility token.',
-    );
+      await Utils.expectRevert(
+        utilityToken.setCoGateway(coGateway.address, { from: owner }),
+        'CoGateway should be linked with this utility token.',
+      );
 
-  });
+    });
```
sarvesh-ost: Mix 4 and 2 space identation.

test/lib/organization/constructor.js
```diff
@@ -18,19 +18,172 @@
...
+    accounts[8],
+  ];
+  let expirationHeight;
+
+  beforeEach(async function () {
+    let currentBlockHeight = await web3.eth.getBlockNumber();
+    expirationHeight = currentBlockHeight + 10;
+  });
+
+  it('should construct with all parameters given', async () => {
```
sarvesh-ost: This should also assert `workers` and it's `expirationHeight`.

test/gateway/eip20_cogateway/confirm_stake_intent.js
schemar: Thanks, fixed 👍

test/gateway/utility_token/set_co_gateway.js
schemar: Thanks, fixed 👍

test/lib/organization/constructor.js
schemar: See `it('should register all given worker addresses with the given expiration height')` lines 94 ff.

PR #557, opened 2018-12-17
Update CHANGELOG.md

schemar:


schemar:
I will add them, but this way it will never me mergeable 😐

sarvesh-ost:
> I will add them, but this way it will never me mergeable 😐

If you can quickly add these please, I will merge right away. 😃

PR #561, opened 2018-12-19
Testing of EIP20CoGateway. confirmStakeIntent()

deepesh-kn:
fixes: #432 
- This PR includes the tests for confirmStakeIntent() functions.
- This PR uses actual merkle proof data for testing.
- Using MockMessageBus for different tests was introducing code duplication and manual effort to sync the files to be consistent with MessageBus.


deepesh-kn:
> Just one last request. It would be great if you could add test case for below logic.
>
>         require(
>             rewardAmount_ <= stakeAmount_,
>             "Reward amount must not be greater than the stake amount."
>         );

That should be handled in a new ticket. #568
As this PR is related to `EIP20CoGateway. confirmStakeIntent()`

test/gateway/eip20_cogateway/confirm_stake_intent.js
```diff
@@ -0,0 +1,552 @@
...
+
+    // Set CoGateway address in mocked utility token.
+    await utilityToken.setCoGatewayAddress(coGateway.address);
+
+    // Set the storage root.
+    await coGateway.setStorageRoot(blockHeight, storageRoot);
+
+  });
+
+  it('should fail when stake address is zero.', async function () {
```
schemar:
```suggestion
  it('should fail when staker address is zero.', async function () {
```

PR #567, opened 2018-12-21
Removed multiple deletes and few code cleanup

deepesh-kn:
Some cleanups.
- Removed the deletion of stake object from stakes mapping form  `stake` as its already done in `progressStakeInternal`.
- Similarly, removed the deletion of redeem object from redeems mapping form `redeem` as its already done in `progressRedeemInternal`.
- Deleted `hasher.sol` it was an old file and was needed anymore.
- Some indentation fixes.


contracts/gateway/EIP20CoGateway.sol
```diff
@@ -801,13 +798,14 @@ contract EIP20CoGateway is GatewayBase {
...
-        bytes32 intentHash = hashStakeIntent(
+        bytes32 intentHash = GatewayLib.hashStakeIntent(
             _amount,
             _beneficiary,
             _staker,
             _stakerNonce,
             _gasPrice,
-            _gasLimit
+            _gasLimit,
+            address(valueToken)
```
sarvesh-ost: Do we need to explicitly add `address` type?

contracts/gateway/EIP20CoGateway.sol
```diff
@@ -928,16 +925,13 @@ contract EIP20CoGateway is GatewayBase {
             _gasLimit
         );
 
-        // Get previousMessageHash.
-        bytes32 previousMessageHash = registerOutboxProcess(
+        // Register new redeem process
```
sarvesh-ost:
```suggestion
        // Register new redeem process.
```

contracts/gateway/EIP20CoGateway.sol
```diff
@@ -928,16 +925,13 @@ contract EIP20CoGateway is GatewayBase {
...
+        registerOutboxProcess(
```
sarvesh-ost: return `previousMessageHash` is not required anymore, it shouldn't be returned from the function

contracts/gateway/EIP20Gateway.sol
```diff
@@ -333,16 +333,13 @@ contract EIP20Gateway is GatewayBase {
             _gasLimit
         );
 
-        // Get previousMessageHash
-        bytes32 previousMessageHash = registerOutboxProcess(
+        // Register for new stake process
```
sarvesh-ost:
```suggestion
        // Register new stake process.
```

Issue #595, opened 2019-01-16
Add getter function to get latest message hash for an address.


deepesh-kn:
Story:

- As a staker, I want to know the status of my last stake/redeem flow.
- By providing the address, I should be able to get message hash and its status.

PR #601, opened 2019-01-18
Add function to get active message hash for a given account address.

deepesh-kn:
- Added `getInboxActiveProcess()` function.
- Added `getOutboxActiveProcess()` function.
- Added corresponding test cases.

fixes: #595


contracts/test/gateway/TestGatewayBase.sol
```diff
@@ -0,0 +1,179 @@
...
+     * @param _messageHash MessageHash for which status is the be set.
+     */
+    function setInboxProcess(
+        address _account,
+        uint256 _nonce,
+        bytes32 _messageHash
+    )
+        external
+    {
+        super.registerInboxProcess(_account, _nonce, _messageHash);
```
sarvesh-ost: Why it's important to have nonce in the argument? 
can we not directly set this?  
```
inboxActiveProcess[_account] = _messageHash;
```

Similar comment for setOutboxProcess

contracts/gateway/GatewayBase.sol
```diff
@@ -338,20 +338,71 @@ contract GatewayBase is Organized {
         emit BountyChangeConfirmed(previousBountyAmount_, changedBountyAmount_);
     }
 
+    /**
+     * @notice Method to get the active message hash and its status from inbox
+     *         for the given account address.
+     *
+     * @dev If message hash do not exists for the given account address then
```
schemar:
```suggestion
     * @dev If message hash does not exist for the given account address then
```
Also, is this not relevant to the user of the function? Why in `@dev`?

contracts/gateway/GatewayBase.sol
```diff
@@ -338,20 +338,71 @@ contract GatewayBase is Organized {
...
+    {
+        messageHash_ = inboxActiveProcess[_account];
+        status_ = messageBox.inbox[messageHash_];
+    }
+
+    /**
+     * @notice Method to get the active message hash and its status from outbox
+     *         for the given account address.
+     *
+     * @dev If message hash do not exists for the given account address then
```
schemar:
```suggestion
     * @dev If message hash does not exist for the given account address then
```
Also, is this not relevant to the user of the function? Why in @dev?

contracts/gateway/GatewayBase.sol
deepesh-kn: Removed from the @dev and added to notice.

PR #620, opened 2019-01-24
Changing state before transfering base token

### LINKED ISSUE #605: Improve code pattern `EIP20CoGateway.progressRedeemInternal`
In `EIP20CoGateway.progressRedeemInternal` the `bounty` is returned to the successful facilitator (`msg.sender`) progressing the redemption (with proof or hashlock).  This `(msg.sender).transfer` is an external call, and a re-entrancy attack is of concern.

Two external functions, `progressRedeem` and `progressRedeemWithProof` check and update the state before calling `progressRedeemInternal`, so a re-entrancy attack is not possible.  However, within `progressRedeemInternal` the `redeems` mapping entry is deleted after the `.transfer`.

It is better to delete the mapping first and check on existence of the entry, before calling `.transfer`

https://github.com/OpenSTFoundation/mosaic-contracts/blob/develop/contracts/gateway/EIP20CoGateway.sol#L1130-L1134

ℹ️ Check other places where the base token is transferred.
- EIP20Gateway.progressStakeInternal
- EIP20Gateway.progressRevertStake
- EIP20Gateway.progressUnstakeInternal
- EIP20CoGateway.progressRevertRedeem
- EIP20CoGateway.progressRedeemInternal
- EIP20CoGateway.progressMintInternal
- OSTPrime.unwrap
- OSTPrime.increaseSupply

sarvesh-ost:
Fixes #605  

State change operations are done before token transfers in order to enhance defense against reentrancy attack.

⚠️ Slither still identifies some vulnerability even after fixing them. It may be the default behavior of slither or there is something which needs to be fixed. In both scenarios, it's out of the scope of this ticket. Any input on this is appreciated.  :pray::skin-tone-3:

```
EIP20CoGateway.progressRevertRedeem (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#641-715) sends eth to arbirary user
	Dangerous calls:
	- burner.transfer(bounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#703)
	- burner.transfer(penalty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#706)
Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
INFO:Detectors:
EIP20CoGateway.progressRedeemInternal (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1116-1150) sends eth to arbirary user
	Dangerous calls:
	- msg.sender.transfer(stakedBounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1139)
Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
```


schemar:
> Fixes #605
> 
> State change operations are done before token transfers in order to enhance defense against reentrancy attack.
> 
> ⚠️ Slither still identifies some vulnerability even after fixing them. It may be the default behavior of slither or there is something which needs to be fixed. In both scenarios, it's out of the scope of this ticket. Any input on this is appreciated. 🙏:skin-tone-3:
> 
> ```
> EIP20CoGateway.progressRevertRedeem (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#641-715) sends eth to arbirary user
> 	Dangerous calls:
> 	- burner.transfer(bounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#703)
> 	- burner.transfer(penalty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#706)
> Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> INFO:Detectors:
> EIP20CoGateway.progressRedeemInternal (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1116-1150) sends eth to arbirary user
> 	Dangerous calls:
> 	- msg.sender.transfer(stakedBounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1139)
> Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> ```

Regarding the first one: I would say it is a false-positive, as the `burner` address is not arbitrary. It is set at construction and cannot be changed.

Regarding the second one: I would say that is a false-positive as well. The function `progressRedeemInternal` is only called from `progressRedeem` and `progressRedeemWithProof`.
For `progressRedeem` the secret must be known. For `progressRedeemWithProof` it is correct that it can be called by an arbitrary address.
However, strictly speaking we *do* have a front-running issue with revealing the secret ...

Anyways, when slither reports something it means that it must be checked by a human, because slither doesn't know for sure. In these case, for example, confidence was "medium".

contracts/gateway/EIP20Gateway.sol
```diff
@@ -617,21 +615,21 @@ contract EIP20Gateway is GatewayBase {
...
+        uint256 stakeBounty = stakes[_messageHash].bounty;
 
-        // transfer the staked amount to the staker
+        delete stakes[_messageHash];
+
+        // Transfer the staked amount to the staker.
         token.transfer(message.sender, amount_);
 
-        // burn facilitator bounty
-        baseToken.transfer(burner, bounty);
```
schemar: This change is hard for me to follow. Where was the original `bounty` taken from??

contracts/gateway/EIP20Gateway.sol
```diff
@@ -617,21 +615,21 @@ contract EIP20Gateway is GatewayBase {
...
-        //penalty charged to staker
-        uint256 penalty = penaltyFromBounty(stakes[_messageHash].bounty);
+        // Burn facilitator bounty.
+        baseToken.transfer(burner, stakeBounty);
+        // Penalty charged to staker.
+        uint256 penalty = penaltyFromBounty(stakeBounty);
```
schemar: Will `penalty` always be zero for a zero bounty? If we cannot guarantee that, the current change could still lead to problems for a zero `stakeBounty` (after deleting the stake from the mapping and reentrying).

contracts/gateway/EIP20Gateway.sol
```diff
@@ -1084,13 +1075,14 @@ contract EIP20Gateway is GatewayBase {
         //Get the stake amount.
         stakeAmount_ = stakes[_messageHash].amount;
 
+        uint256 stakedBounty = stakes[_messageHash].bounty;
+
+        delete stakes[_messageHash];
```
schemar: This deletion does not prevent the following `token.transfer(address(stakeVault), stakeAmount_);` in case of reentrancy. Problem?

contracts/gateway/EIP20Gateway.sol
```diff
@@ -1159,8 +1151,13 @@ contract EIP20Gateway is GatewayBase {
         );
 
         unstakeAmount_ = redeemAmount_.sub(rewardAmount_);
+
+        address beneficiary = unstakes[_messageHash].beneficiary;
+
+        delete unstakes[_messageHash];
```
schemar: This delete does not prevent the execution of the next line `stakeVault.releaseTo(beneficiary, unstakeAmount_);`...

contracts/gateway/EIP20CoGateway.sol
```diff
@@ -1053,6 +1054,8 @@ contract EIP20CoGateway is GatewayBase {
         beneficiary_ = mint.beneficiary;
         stakeAmount_ = mint.amount;
 
+        address payable beneficiary = mint.beneficiary;
```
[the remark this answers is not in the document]
sarvesh-ost: Yes, eth transfer can't be done for `beneficiary_` but for `beneficiary`

contracts/gateway/EIP20CoGateway.sol
```diff
@@ -1067,9 +1070,11 @@ contract EIP20CoGateway is GatewayBase {
 
         mintedAmount_ = stakeAmount_.sub(rewardAmount_);
 
+        delete mints[_messageHash];
```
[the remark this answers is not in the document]
sarvesh-ost: If you see L296, it wouldn't allow reentrancy at the first place(Due to message bus state machine).  This is true for all the methods in the scope of this PR. 

However, do you feel there should be an explicit check for below condition?
 
```
require(
            mints[_messageHash].amount > 0,
            "Mint amout must be greater than zero."
        );
```

contracts/gateway/EIP20Gateway.sol
sarvesh-ost: Reentrancy will be handled by `progressOutboxRevocation` on L606. What do you suggest?
Somethin like this?
require(
            stakes[_messageHash].amount > 0,
            "Stake amount must be greater than zero."
)

sarvesh-ost:
> > Fixes #605
> > State change operations are done before token transfers in order to enhance defense against reentrancy attack.
> > ⚠️ Slither still identifies some vulnerability even after fixing them. It may be the default behavior of slither or there is something which needs to be fixed. In both scenarios, it's out of the scope of this ticket. Any input on this is appreciated. 🙏:skin-tone-3:
> > ```
> > EIP20CoGateway.progressRevertRedeem (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#641-715) sends eth to arbirary user
> > 	Dangerous calls:
> > 	- burner.transfer(bounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#703)
> > 	- burner.transfer(penalty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#706)
> > Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> > INFO:Detectors:
> > EIP20CoGateway.progressRedeemInternal (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1116-1150) sends eth to arbirary user
> > 	Dangerous calls:
> > 	- msg.sender.transfer(stakedBounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1139)
> > Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> > ```
> 
> Regarding the first one: I would say it is a false-positive, as the `burner` address is not arbitrary. It is set at construction and cannot be changed.
> 
> Regarding the second one: I would say that is a false-positive as well. The function `progressRedeemInternal` is only called from `progressRedeem` and `progressRedeemWithProof`.
> For `progressRedeem` the secret must be known. For `progressRedeemWithProof` it is correct that it can be called by an arbitrary address.
> However, strictly speaking we _do_ have a front-running issue with revealing the secret ...
> 
> Anyways, when slither reports something it means that it must be checked by a human, because slither doesn't know for sure. In these cases, for example, confidence was "medium".

Thanks for the review and detailed feedback. 🙌  

As you have observed, Message bus state machine will always protect against reentrancy. This deletion is like another safety measure following the general practice of changing state before token transfer. 

Do you think there should be `require` to ensure mapping is not already deleted? 
Happy to discuss more.

contracts/gateway/EIP20CoGateway.sol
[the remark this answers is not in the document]
schemar: Why don't you make `beneficiary_` payable?

schemar:
> > > Fixes #605
> > > State change operations are done before token transfers in order to enhance defense against reentrancy attack.
> > > ⚠️ Slither still identifies some vulnerability even after fixing them. It may be the default behavior of slither or there is something which needs to be fixed. In both scenarios, it's out of the scope of this ticket. Any input on this is appreciated. 🙏:skin-tone-3:
> > > ```
> > > EIP20CoGateway.progressRevertRedeem (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#641-715) sends eth to arbirary user
> > > 	Dangerous calls:
> > > 	- burner.transfer(bounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#703)
> > > 	- burner.transfer(penalty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#706)
> > > Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> > > INFO:Detectors:
> > > EIP20CoGateway.progressRedeemInternal (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1116-1150) sends eth to arbirary user
> > > 	Dangerous calls:
> > > 	- msg.sender.transfer(stakedBounty) (mosaic-contracts/contracts/gateway/EIP20CoGateway.sol#1139)
> > > Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#functions-that-send-ether-to-arbitrary-destinations
> > > ```
> > 
> > 
> > Regarding the first one: I would say it is a false-positive, as the `burner` address is not arbitrary. It is set at construction and cannot be changed.
> > Regarding the second one: I would say that is a false-positive as well. The function `progressRedeemInternal` is only called from `progressRedeem` and `progressRedeemWithProof`.
> > For `progressRedeem` the secret must be known. For `progressRedeemWithProof` it is correct that it can be called by an arbitrary address.
> > However, strictly speaking we _do_ have a front-running issue with revealing the secret ...
> > Anyways, when slither reports something it means that it must be checked by a human, because slither doesn't know for sure. In these cases, for example, confidence was "medium".
> 
> Thanks for the review and detailed feedback. 🙌
> 
> As you have observed, Message bus state machine will always protect against reentrancy. This deletion is like another safety measure following the general practice of changing state before token transfer.
> 
> Do you think there should be `require` to ensure mapping is not already deleted?
> Happy to discuss more.

Ah yes. You are right. I only focused on the mappings and didn't regard the state machine. Let me resolve what I can 😊

schemar:
OK now I see. It was "invisible" to me due to it being handled in the message bus and it relies on calling the right methods on the message bus in the right order.
Overall the Gateway/MessageBus architecture is very opaque and probably also not so easy to understand by the people that would do a security assessment.
@sarvesh-ost should we add to the `@dev` section of the affected functions that correct order of execution is handled by the message bus state machine? What do you think?

contracts/gateway/EIP20Gateway.sol
schemar: I cannot find the variable `bounty` in the original code 😳

contracts/gateway/EIP20CoGateway.sol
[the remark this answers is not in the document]
sarvesh-ost: Renamed to payableBeneficiary.

sarvesh-ost:
> Looks good 🐨
> 
> Only `beneficiary`/`beneficiary_`. I think one is sufficient.
> And the question whether to add comments about the state machine.

@schemar  Do you still want me to add dev comment about execution sequence handled by message bus, as we have explicit checks to verify request exists(it's no more dependent on message bus)

If yes, Do you have a suggestion?

PR #624, opened 2019-01-28
Integration tests can access deployed contracts

### LINKED ISSUE #608: Set-up integration test environment
The integration test environment should consist of two chains, one geth node each. @hobofan can advise in how to easily get that up and running.

Then, the integration tests should use "chain deployment" to deploy the contracts.

Mocha and Chai should be used to test the functionality.

Integration tests should be run with a separate npm command.

Refer:
- https://github.com/OpenSTFoundation/chains-utils
- https://hub.docker.com/r/ethereum/client-go

schemar:
The main file is `test_integration/main.sh` as called from
`package.json`.

Docker compose spawns two instances of geth, one for origin and one for
auxiliary. It also tears them down after the tests.

In between the tests are run. `integration_tests.js` puts everything
together (truffle and mocha) and executes the tests in the integration
test directory.

Deployment is required to deploy the contracts and add them to the
`shared` module.

The `shared` module is an object that stores data across tests. For
example Web3 instances or TruffleContract instances.

Stake and mint is just an example.

As examples I added two Anchors and the two Gateways to this PR. Can be
extended as required.

I had to use `truffle exec` in order to have access to
`artifacts.require`.

Fixes #608


test_integration/integration_tests.js
```diff
@@ -0,0 +1,61 @@
...
+        enableTimeouts: false,
+    });
+
+    Mocha.utils.lookupFiles(__dirname, ['js'], true)
+        .filter(
+            // Skipping this file so that artifacts are not loaded.
+            file => file.substr(-20) !== 'integration_tests.js',
+        )
+        .forEach((file) => {
+            mocha.addFile(file);
```
sarvesh-ost: I think this logic doesn't take folder name in consideration. This might result in the wrong ordering of execution. 🤔

schemar: I *assume* the util method to get the files does.

schemar: I tested it by renaming the `stake` file and it still executes in the right order.

test_integration/shared.js
```diff
@@ -0,0 +1,89 @@
...
+ *     contract name, e.g. `EIP20Gateway` or `Anchor`. Exception: `BaseToken`
+ *     and `BrandedToken`.
+ * @property {string} deployerAddress The address of the account that deployed
+ *     the contracts in the deployment phase.
+ */
+class Chain {
+    constructor() {
+        this.web3 = {};
+        this.contractAddresses = {};
+        this.contracts = {};
```
sarvesh-ost: This should probably have accounts of that chain.

tools/deployment_tool/utils.js
```diff
@@ -154,7 +155,8 @@ const deployContracts = async (
             gas = await web3.eth.estimateGas(deploymentObject.transactionObject);
         }
         if (!gasPrice) {
-            gasPrice = await web3.eth.getGasPrice();
+            gasPrice = new BN(await web3.eth.getGasPrice());
+            gasPrice = gasPrice.muln(10).toString(10);
```
sarvesh-ost: I didn't understand this part 🤔

truffle.js
```diff
@@ -1,38 +1,41 @@
...
-      network_id: "*",
-      port: 8555,         // <-- If you change this, also set the port option in .solcover.js.
-      gas: 0xfffffffffff, // <-- Use this high gas value 
-      gasPrice: 0x01,     // <-- Use this low gas price
+    networks: {
+        development: {
+            host: 'localhost',
+            network_id: '*', // Match any network id
+            port: 8545,
+            gas: 12000000,
```
sarvesh-ost: This should be changed to 8 million. For an example, if contract deployment cost exceeds 8 million, the unit test should fail 💭  
8 million is block gas limit on main-net and rospten.

test_integration/shared.js
```diff
@@ -0,0 +1,89 @@
...
+ * @property {Object} contractAddresses All addresses in sting format, indexed
+ *     by the contract name, as written in the solidity source file.
+ * @property {Object} contracts Instances of TruffleContracts. Indexed by their
+ *     contract name, e.g. `EIP20Gateway` or `Anchor`. Exception: `BaseToken`
+ *     and `BrandedToken`.
+ * @property {string} deployerAddress The address of the account that deployed
+ *     the contracts in the deployment phase.
+ */
+class Chain {
+    constructor() {
```
sarvesh-ost: How integration test will know about owner of organization for the transaction like `anchorStateRoot`?

test_integration/01_deployment/01_deploy.js
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 OpenST Ltd.
+// Copyright 2018 OpenST Ltd.
```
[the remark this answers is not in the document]
schemar: Fixed

test_integration/01_deployment/01_deploy.js
```diff
@@ -33,49 +31,71 @@ const {
...
+            deployerAddressOrigin,
+            'new',
+        );
         assert(
             Web3.utils.isAddress(baseTokenAddressOrigin),
             'Did not correctly deploy base token on Origin.',
         );
+
+        /* Note that they are called BrandedToken and BaseToken! */
+        shared.origin.contractAddresses.BrandedToken = tokenAddressOrigin;
```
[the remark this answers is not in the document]
schemar: How do you want to call it to differentiate it from the base token?

test_integration/01_deployment/01_deploy.js
```diff
@@ -33,49 +31,71 @@ const {
...
+        shared.origin.deployerAddress = deployerAddressOrigin;
+        shared.auxiliary.deployerAddress = deployerAddressAuxiliary;
     });
 
-    after(() => {
-        dockerTeardown();
+    after(async () => {
+        await shared.origin.addContract('EIP20Gateway');
+        await shared.origin.addContract('Anchor');
+        await shared.origin.addContract('EIP20StandardToken', 'BrandedToken');
```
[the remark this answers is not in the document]
schemar: Added

test_integration/shared.js
schemar: `@property {string} deployerAddress`

test_integration/shared.js
schemar: You can get the accounts from `web3`

tools/deployment_tool/utils.js
schemar: I deleted it. We had to increase the gas price for testing in order for our transactions to be mined timely.

truffle.js
schemar: I don't think that is in the scope of this ticket 🤔 I am afraid to break stuff.

test_integration/integration_tests.js
```diff
@@ -0,0 +1,64 @@
...
+const EIP20Gateway = artifacts.require('EIP20Gateway');
+const EIP20CoGateway = artifacts.require('EIP20CoGateway');
+const EIP20StandardToken = artifacts.require('EIP20StandardToken');
+
+const setupArtifacts = () => {
+    shared.artifacts = {
+        Anchor,
+        EIP20Gateway,
+        EIP20CoGateway,
+        EIP20StandardToken,
```
deepesh-kn: We have more contracts that will be needed for testing. Why `OSTPrime`, `MessageBus`,`GatewayLib` ... etc are not included in this?

test_integration/integration_tests.js
```diff
@@ -0,0 +1,64 @@
...
+
+const runTests = (callback) => {
+    const mocha = new Mocha({
+        enableTimeouts: false,
+    });
+
+    Mocha.utils.lookupFiles(__dirname, ['js'], true)
+        .filter(
+            // Skipping this file so that artifacts are not loaded.
+            file => file.substr(-20) !== 'integration_tests.js',
```
deepesh-kn: Will any other file named like `1_integration_test.js` will be skipped?
```suggestion
            file => file.substr(-21) !== '/integration_tests.js',
```

truffle.js
deepesh-kn: Ideally, it should not break, but If it breaks we will get a chance to fix it, or create a ticket. Atleast we will know.

contracts/test/EIP20StandardToken.sol
```diff
@@ -0,0 +1,61 @@
+pragma solidity ^0.5.0;
+
+// Copyright 2018 OpenST Ltd.
```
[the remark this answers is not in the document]
schemar: ✅ fixed

contracts/test/EIP20StandardToken.sol
```diff
@@ -0,0 +1,61 @@
...
+        uint8 _decimals
+    )
+        public
+        EIP20Token(
+            _symbol,
+            _name,
+            _decimals
+        )
+    {
+        totalTokenSupply = _totalSupply * 10**uint256(_decimals);
```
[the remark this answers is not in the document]
schemar: OK I changed it.

test_integration/integration_tests.js
schemar: This is about integration tests. So what we want to do is test the behavior of the system from the consumer's point of view. The consumer interacts with the anchor (anchor state root), the gateways (stake and redeem) and an EIP20StandardToken (approve transfer).
The user would not interact with the message bus or the gateway lib. They are, however, used by the others and should produce correct results, which will be asserted.

About OSTPrime I am not sure. Do we make calls to it directly? E.g. to assert balances? It is right now not part of the deployment process. Should it be added?

test_integration/integration_tests.js
schemar: Good point. Updated to `file => file !== __filename,`

test_integration/shared.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * It is important that every `require` is written exactly `shared`,
+ * case-sensitive!
+ */
+
+/**
+ * Contains all relevant data on a chain.
+ *
+ * @property {Web3} web3 A web3 object to access the chain.
+ * @property {Object} contractAddresses All addresses in sting format, indexed
```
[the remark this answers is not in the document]
schemar: ✅ Check

truffle.js
schemar: OK I updated it. Let's see if CI passes 😱

test_integration/shared.js
```diff
@@ -0,0 +1,89 @@
...
+ *     by the contract name, as written in the solidity source file.
+ * @property {Object} contracts Instances of TruffleContracts. Indexed by their
+ *     contract name, e.g. `EIP20Gateway` or `Anchor`. Exception: `BaseToken`
```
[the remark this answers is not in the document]
schemar: Comment updated 🔨

test_integration/01_deployment/01_deploy.js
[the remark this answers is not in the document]
sarvesh-ost: I think just`token` is fine.

test_integration/shared.js
sarvesh-ost: If it's available with shared, it's very convenient for the integration tests.  Just as truffle provide us with accounts right for unit tests. Similar concept.

schemar: Should I add more things from web3 (and maybe the contract artifacts or instances) to shared as well?

test_integration/shared.js
schemar: Because the owner is, by default in the contract, the deployer. Do you want me to copy the deployer address to a key "organization"?

test_integration/integration_tests.js
schemar: OK 👍 Can we skip it here, then @deepesh-kn? Will be added in @sarvesh-ost's PR.

test_integration/shared.js
schemar: Oops, you are right of course. Sorry. I will for now copy the address to an `organization` key. Unless you have a better idea?

sarvesh-ost: Sounds good to me.

test_integration/shared.js
sarvesh-ost: We already have artifacts and instance(Per chain) in the shared, I believe. Am i missing something here? 🤔

schemar: I mean details from within the artifacts or instances? For easier access from tests.

test_integration/integration_tests.js
deepesh-kn: 👍

test_integration/shared.js
sarvesh-ost: No, I feel other details are test specific. Accounts are must have thing for all the tests.

sarvesh-ost: I realized I have added this in Stake and mint integration test.  We can skip for this PR 👍

PR #626, opened 2019-01-28
Initializing all variables

### LINKED ISSUE #618: Check for possible usage of uninitialized state variables.
Initialize all the variables. 
If a variable is meant to be initialized to zero, explicitly set it to zero.

- GatewayBase.encodedGatewayPath
- GatewayBase.messageBox
- MosaicCore.height
- GatewayBase.remoteGateway

schemar:
Fixes #618


contracts/gateway/GatewayBase.sol
```diff
@@ -171,8 +171,12 @@ contract GatewayBase is Organized {
         );
 
         stateRootProvider = _stateRootProvider;
-
         bounty = _bounty;
+
+        // The following variables are not known at construction:
+        messageBox = MessageBus.MessageBox();
+        encodedGatewayPath = '';
```
sarvesh-ost: ' ' -> 0x ?

sarvesh-ost: In case of EIP20CoGateway, this value is created on construction. 
I hope parent contract constructor is called before the child and doesn't override it. 
This understanding is correct? 

See [here](https://github.com/OpenSTFoundation/mosaic-contracts/blob/0186a2e0e5d8d3780facb2d2c148dcb8edd10d66/contracts/gateway/EIP20CoGateway.sol#L250)

schemar: `0x` is not valid

schemar: base constructors are executed first. that is correct.

PR #627, opened 2019-01-28
Tests assert that the gateway nonce increases by 1

### LINKED ISSUE #625: Add unit tests to assert that nonce for Staker/redeemer  should increase by 1 after successful stake/redeem  in EIP20Gateway and CoGateway
In EIP20Gateway and EIP20Cogateway nonce is maintained. For each successful stake or redeem request, staker/redeemer nonce should be increased by one. 

 Note: This tests can be covered in stake/redeem tests ?

schemar:
I extended the `stake.js` and `redeem.js` tests to include an additional
testcase. It asserts that the nonce increases by 1 when stake or redeem
is called on the (co)gateway.

The alternative to put the test cases into `get_nonce.js` wasn't
feasible. Extracting the stake/redeem functionality from `stake.js` and
`redeem.js` would be very tedious and time consuming. Instead, I added a
comment to the `get_nonce.js` file that the tests about increasing the
nonce are part of the stake and redeem tests.

I updated the affected files to pass linting.

Fixes #625


schemar:
@deepesh-kn / @sarvesh-ost in the changed files, look for these two test cases:
* https://github.com/OpenSTFoundation/mosaic-contracts/pull/627/files#diff-0887b4556a436a3d4fe8ccc496e5c905R516
* https://github.com/OpenSTFoundation/mosaic-contracts/pull/627/files#diff-27c45ec473ccc992f7d075685276adfdR153

PR #651, opened 2019-02-20
Fixed bug: For extension node condition should use current path pointer

### LINKED ISSUE #652: Merkle Proof verification failed for a valid proof.
https://github.com/OpenSTFoundation/mosaic-contracts/blob/c48747a91746061fba73fc7a9189019d7b62fbbe/contracts/lib/MerklePatriciaProof.sol#L20

On[ L63,`_nibblesToTraverse`](https://github.com/OpenSTFoundation/mosaic-contracts/blob/c48747a91746061fba73fc7a9189019d7b62fbbe/contracts/lib/MerklePatriciaProof.sol#L63) function increases pathPtr with the size of the extension node key.

`pathPtr += _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);`

Then on [L73](https://github.com/OpenSTFoundation/mosaic-contracts/blob/c48747a91746061fba73fc7a9189019d7b62fbbe/contracts/lib/MerklePatriciaProof.sol#L73), updated `pathPtr` is passed. Here `_nibblesToTraverse` will return `0` because the key of the extension node will not match this time. Hence `false` is returned from the function and Merkle proof verification will fail.

Possible fix for the issue is to preserve `pathPtr` before updating on L63 and pass the preserved `pathPtr` on L73. Refer this [commit](https://github.com/sarvesh-ost/mosaic-contracts/commit/97714b646ada224f425463128810aae21893c58d).

Below are the parameters which were passed to the `verify` function.

_value_ : 0x5fe7f977e71dba2ea1a68e21057beebb9be2ac30c6410aa38d4f3fbe41dcffd2

_encodedPath_ : 0x40769c9cc6db1aff35e37923897950227d2a87c95ef0101b2a0233740f2ce00d

_rlpParentNodes_ : 0xf9022cf9019180a0320ce86e317f7e61e7890a6671338aa79374e5ec3206fc38496b5b3719788fc4a07642cb37fbbc28c4c2e8367a93bb491013223d266543c0e42db6a580cc274beea0ad8732b76b5befaa3c523633abdafb8b410b8e1140526255c839a7225e2b2f3ea060b926526a0a95873151f96f0c4aa7746dd1cc51242b4893d43554acd2716cda80a03e819a317d72fa7445677a4104b74899547b424aec05ae303b605f7d78aa9dc4a0c9980a15de3258be2f1dcf5d0657c256f7e7aebe83b736aada89aa02dee3853ea0747f61b4309342612ee827a5930678c61bd54a2c85e207144c265912351d46f4a070256c377a27b7c7cdf605559f21f20f54d147abce914c3b1ac36815fd8f2b8080a0b7a25a664618fb58b0754dd3c9665402787704d11169240bccf8e7c34feb0167a0e78c6966deeb91437d0c2cbf66bde64fd4dbb76d22a9885679a3cf9e5fd84394a02a74dc63db6f6f5f79d6cadbb2b348771d3bcd1913f9319cfa0c5fea4353db47a0e53cc96cd3fe0910d63b48800ac824d7f0e094548d259d6342da2189ca3c0e238080e210a0094e521c10bbb9291d38d4aebd6c6722a4763451011f97f24709b28611831699f8518080808080a0460b29d4cbb31e3680769b8ece024e6e35e9c9c3d9c7c207873e93b8c43cff1980a0fe9627f9a9fb3323465824fc6fdfd87f907332a17f6ed39bfdcd94c76b548eee808080808080808080e19f369c9cc6db1aff35e37923897950227d2a87c95ef0101b2a0233740f2ce00d01

_root_ : 0xa506e7e444fc636360bb593a808f41f126a8813ea71410d4515987eacc594380

sarvesh-ost:
Fixes #652


contracts/lib/MerklePatriciaProof.sol
```diff
@@ -70,7 +71,7 @@ library MerklePatriciaProof {
                     }
                 }
                 //extension node ... test if means that it is empty value
-                if(_nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr) == 0) {
+                if(_nibblesToTraverse(RLP.toData(currentNodeList[0]), path, currentpathPtr) == 0) {
                     return (keccak256(abi.encodePacked()) == value);
```
schemar: it should also just `return false`

contracts/lib/MerklePatriciaProof.sol
```diff
@@ -60,6 +60,7 @@ library MerklePatriciaProof {
                 nodeKey = RLP.toBytes32(currentNodeList[nextPathNibble]);
                 pathPtr += 1;
             } else if(currentNodeList.length == 2) {
+                uint currentpathPtr = pathPtr;
```
schemar: Wouldn't it be more intuitive to keep the `pathPtr` and instead introduce a new variable for the leaf node check? (e.g. `advancedPathPointer` or something with a better name.

contracts/lib/MerklePatriciaProof.sol
```diff
@@ -70,7 +71,7 @@ library MerklePatriciaProof {
                     }
                 }
                 //extension node ... test if means that it is empty value
```
schemar: This comment should also be changed to make more sense. It is just an extension node that should traverse more than 0 nibbles.

contracts/lib/MerklePatriciaProof.sol
```diff
@@ -60,6 +60,7 @@ library MerklePatriciaProof {
                 nodeKey = RLP.toBytes32(currentNodeList[nextPathNibble]);
                 pathPtr += 1;
             } else if(currentNodeList.length == 2) {
+                uint currentpathPtr = pathPtr;
                 pathPtr += _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);
```
benjaminbollen: we potentially calculate `_nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr)` twice.

suggestion:
```
uint traverseLength = _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);
if(pathPtr + traverseLength == path.length) { //leaf node
            if(keccak256(abi.encodePacked(RLP.toData(currentNodeList[1]))) == value) {
                   return true;
            } else {
                   return false;
            }
} else if (traverseLength == 0) { // error: couldn't traverse path
           return false
} else { // extension node
           pathPtr += traverseLength;
}

File: contracts/lib/MerklePatriciaProof.sol (hunk 1 above)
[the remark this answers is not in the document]
Comment (benjaminbollen) [reply contracts/lib/MerklePatriciaProof.sol]: see comment above for "my understanding" of the logic, plus rewrite

File: contracts/lib/MerklePatriciaProof.sol (hunk 2 above)
Comment (sarvesh-ost) [reply contracts/lib/MerklePatriciaProof.sol]: Would that not be confusing ? 🤔 
It would look like this. In below snippet, again `pathPtr` needs to be updated with `advancedPathPointer`

```
                uint advancedPathPointer = pathPtr;
                advancedPathPointer += _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);

                if(pathPtr == path.length) {//leaf node
                    if(keccak256(abi.encodePacked(RLP.toData(currentNodeList[1]))) == value) {
                        return true;
                    } else {
                        return false;
                    }
                }
                //extension node ... test if means that it is empty value
                if(_nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr) == 0) {
                    return false;
                }
                pathPtr = advancedPathPointer;
```

File: contracts/lib/MerklePatriciaProof.sol (hunk 4 above)
Comment (schemar) [reply contracts/lib/MerklePatriciaProof.sol]: The last `else` block could also include the `nodeKey = RLP.toBytes32(currentNodeList[1]);` part.

**Comment by schemar** [comment]
I just want to restate that we also need a lot of positive and negative test cases. Can we add some? Generally, any bug fix should come with at least the unit test that fails due to the bug and passes after the fix.

**Comment by sarvesh-ost** [comment]
> I just want to restate that we also need a lot of positive and negative test cases. Can we add some? Generally, any bug fix should come with at least the unit test that fails due to the bug and passes after the fix.

I can easily add tests for the cases which kit reported.  👍

File: contracts/lib/MerklePatriciaProof.sol (hunk 4 above)
Comment (sarvesh-ost) [reply contracts/lib/MerklePatriciaProof.sol]: > we potentially calculate `_nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr)` twice.
> 
> suggestion:
> 
> ```
> uint traverseLength = _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);
> if(pathPtr + traverseLength == path.length) { //leaf node
>             if(keccak256(abi.encodePacked(RLP.toData(currentNodeList[1]))) == value) {
>                    return true;
>             } else {
>                    return false;
>             }
> } else if (traverseLength == 0) { // error: couldn't traverse path
>            return false
> } else { // extension node
>            pathPtr += traverseLength;
> }
> ```

Are you fine with this? In the original code, pathPtr was always updated, Can we also keep like that? 
Though, your suggestion looks absolutely correct as per the specification. 
```
uint traverseLength = _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);
if(pathPtr + traverseLength == path.length) { //leaf node
            if(keccak256(abi.encodePacked(RLP.toData(currentNodeList[1]))) == value) {
                   return true;
            } else {
                   return false;
            }
} else if (traverseLength == 0) { // error: couldn't traverse path
           return false
} 
 pathPtr += traverseLength;

```

Comment (schemar) [reply contracts/lib/MerklePatriciaProof.sol]: I think Ben's addition of the explicit `else` improves readability.

File: contracts/lib/MerklePatriciaProof.sol (hunk 2 above)
Comment (schemar) [reply contracts/lib/MerklePatriciaProof.sol]: "advanced path pointer" would be required in the leaf node check and the 0 length check. Anyways, I prefer Ben's suggestion, actually.

File: contracts/lib/MerklePatriciaProof.sol (hunk 1 above)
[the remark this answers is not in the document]
Comment (sarvesh-ost) [reply contracts/lib/MerklePatriciaProof.sol]: I agree with the reasoning 👍

File: contracts/lib/MerklePatriciaProof.sol [hunk 5]
```diff
@@ -59,22 +59,24 @@ library MerklePatriciaProof {
                 if(nextPathNibble > 16) {return false;}
                 nodeKey = RLP.toBytes32(currentNodeList[nextPathNibble]);
                 pathPtr += 1;
-            } else if(currentNodeList.length == 2) {
-                pathPtr += _nibblesToTraverse(RLP.toData(currentNodeList[0]), path, pathPtr);
+            }else if(currentNodeList.length == 2) {
```
schemar:
```suggestion
            } else if(currentNodeList.length == 2) {
```

contracts/lib/MerklePatriciaProof.sol
```diff
@@ -59,22 +59,24 @@ library MerklePatriciaProof {
...
 
-                if(pathPtr == path.length) {//leaf node
+                // Count of matched bytes in path starting from pathPtr and node key.
```
schemar:
```suggestion
                // Count of matching node key nibbles in path starting from pathPtr.
```

test/lib/merkle_patricia_proof.js
```diff
@@ -1,19 +1,37 @@
...
+// ----------------------------------------------------------------------------
+
+const MerklePatriciaProof = artifacts.require('./MerklePatriciaProof.sol');
 const proofJson = require('../data/proof.json');
 const Utils = require('../test_lib/utils.js');
 
-contract('MerklePatriciaProof', (accounts) => {
-  let merkleMock = null;
+contract('MerklePatriciaProof.verify()', (accounts) => {
+  let merklePatriciaLib = null;
```
schemar:
```suggestion
  let merklePatriciaLib;
```

benjaminbollen:
I propose two paths:
1. we can keep this hotfix on the branch, and over the coming days we put effort to adding more explicit unit tests
2. we merge it with `release-0.10`, `develop` and publish `beta.3` and we put effort to add more explicit unit tests 😝

sarvesh-ost:
> Nice! However, can we also add tests where verification should fail and assert that they still fail with the new code? I am not sure how much we want to test this, but from my point of view 2 tests seems a little slim.
> 
> Also minor in-line comments.
@schemar These two tests are the scenarios where the bug was reported. It makes sense to add these tests that give confidence that the fix works.

However, what you are requesting is to test the MPP library in general, which I would say should be done as a separate ticket. Testing MPP library for all the scenarios is a significant effort, as we don't have proof for all the possible trie trees. These need to be manually formed(Unless there is a better alternative) 

I am not sure if we want to do this for this fix. 
@benjaminbollen  what's your call?

benjaminbollen:
@sarvesh-ost yes, I think this hotfix PR should go ahead so that we can test it, both in KIT and in JLP examples;

to @schemar s point, yes let's put focus on additional deliberate tests, but this is more work than this issue.

I propose we merge this into the hot fix branch; *and* go forward with publishing `beta.3` so we can test it from KIT and JLP

PR #669, opened 2019-03-14
GatewayBase uses circular buffer for storing storage roots.

pgev:


benjaminbollen:
decided to move this optimization to the next version.  closing PR

PR #670, opened 2019-03-14
Added UT.exists() 

deepesh-kn:
This PR includes the following changes
- Added function `exists()` in `UtilityTokenInterface.sol`
- Function `exists()` returns `true` for all addresses in `UtilityToken.sol`
- Added a require to check if the `beneficiary` address exists in utility token
- Updated the `confirmStakeIntent` unit test 
- Added unit test for exists() function. [Reviewer may check if this is required or not]


contracts/gateway/EIP20CoGateway.sol
```diff
@@ -775,6 +775,10 @@ contract EIP20CoGateway is GatewayBase {
             _rlpParentNodes.length != 0,
             "RLP parent nodes must not be zero."
         );
+        require(
+            UtilityTokenInterface(utilityToken).exists(_beneficiary),
+            "Beneficiary address must be a registered address in utility token."
```
sarvesh-ost: We don't register addresses at the mosaic layer. Maybe, a more general message like 
`Beneficiary address must be allowed to hold utility token.`

contracts/gateway/UtilityToken.sol
```diff
@@ -180,6 +180,19 @@ contract UtilityToken is EIP20Token, Organized, UtilityTokenInterface {
         success_ = decreaseSupplyInternal(_amount);
     }
 
+    /**
+     * @notice Checks if the given address is registered in the economy
```
sarvesh-ost: Similar comment on language.

contracts/gateway/UtilityToken.sol
```diff
@@ -180,6 +180,19 @@ contract UtilityToken is EIP20Token, Organized, UtilityTokenInterface {
...
+     *
+     * @dev In case of internal economy, an address has to be registered.
+     *      This utility token does not have any internal actors so it returns
+     *      true for all addresses.
+     *
+     * @return exists_ `true` for all given address
+     */
+    function exists(address) external returns (bool exists_) {
```
sarvesh-ost: This can be a pure function, I think.

contracts/test/gateway/MockUtilityToken.sol
```diff
@@ -70,4 +86,15 @@ contract MockUtilityToken is UtilityToken {
...
         coGateway = _coGatewayAddress;
     }
+
+    /**
+     * @notice Set the addressExists variable to mock return value of
+     *         function exists(). This is only for testing.
+     *
+     * @param _exists Boolean value
+     */
+    function mockAddressExists(bool _exists) external {
```
sarvesh-ost: An address is never mocked, instances/methods are mocked 🤔

test/gateway/utility_token/exists.js
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+  });
+
+  it('should always return true for any given address', async () => {
+    let account5ExistsResult = await utilityToken.exists.call(accounts[5]);
+    assert.strictEqual(
+      account5ExistsResult,
+      true,
+      'Contract should return true.'
```
sarvesh-ost:
```suggestion
      'Contract should return true.',
```

test/gateway/utility_token/exists.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+
+    let account6ExistsResult = await utilityToken.exists.call(accounts[6]);
+    assert.strictEqual(
+      account6ExistsResult,
+      true,
+      'Contract should return true.'
```
sarvesh-ost:
```suggestion
      'Contract should return true.',
```

test/gateway/utility_token/exists.js
```diff
@@ -0,0 +1,68 @@
...
+      TOKEN_SYMBOL,
+      TOKEN_NAME,
+      TOKEN_DECIMALS,
+      organization.address,
+    )
+
+  });
+
+  it('should always return true for any given address', async () => {
+    let account5ExistsResult = await utilityToken.exists.call(accounts[5]);
```
sarvesh-ost:
```suggestion
    const account5ExistsResult = await utilityToken.exists.call(accounts[5]);
```

test/gateway/utility_token/exists.js
```diff
@@ -0,0 +1,68 @@
...
+    assert.strictEqual(
+      account5ExistsResult,
+      true,
+      'Contract should return true.'
+    );
+
+    let account6ExistsResult = await utilityToken.exists.call(accounts[6]);
```
sarvesh-ost:
```suggestion
    const account6ExistsResult = await utilityToken.exists.call(accounts[6]);
```

contracts/gateway/UtilityToken.sol
deepesh-kn: This can be pure for this particular case, but other UtilityToken implementations may not be pure. I am not sure if it will work in that case.

contracts/gateway/EIP20CoGateway.sol
```diff
@@ -775,6 +775,10 @@ contract EIP20CoGateway is GatewayBase {
             _rlpParentNodes.length != 0,
             "RLP parent nodes must not be zero."
         );
+        require(
+            UtilityTokenInterface(utilityToken).exists(_beneficiary),
+            "Beneficiary address must be allowed to hold utility token."
```
benjaminbollen: "beneficiary address must exist."

contracts/gateway/UtilityToken.sol
```diff
@@ -180,6 +180,18 @@ contract UtilityToken is EIP20Token, Organized, UtilityTokenInterface {
         success_ = decreaseSupplyInternal(_amount);
     }
 
+    /**
+     * @notice Checks if an address is allowed to hold utility token.
+     *
+     * @dev In this case all the addresses are allowed to hold the utility
```
benjaminbollen: dev. For standard ethereum all account addresses exist by default, so it returns true for all addresses.

benjaminbollen:
otherwise, looks great! 👍

Issue #662, opened 2019-03-13
Gateway contracts feedbacks 


sarvesh-ost:
1. Removed value token (Please double check the usage) 
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20CoGateway.sol#L216

2. Rename `redeemProcess` -> redeem
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20CoGateway.sol#L722

3. Remove the nonce parameter from EIP20 Gateway `stake` and EIP20CoGateway `redeem` method [Not 0.0.10 release]

4. Allow zero amount in `increaseSupply`. Also, Check UT if allows zero transfer. 
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20CoGateway.sol#L1139

5. Try to make events specific to message bus states. The intention is to reduce the number of event types from the gateway.
 [Not 0.0.10 release]

6. Don't return penalty in case of progress with proof. Burn the penalty. Check progressStakeWithProof and progressRedeemWithProof

https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20CoGateway.sol#L569

https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20Gateway.sol#L500

7. Move deletion up 
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20Gateway.sol#L1011

https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20CoGateway.sol#L451

8. Emit event on activating and deactivate Gateway 
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/EIP20Gateway.sol#L1046

9. Move check that reward > stake/mint amount in confirmStakeIntent and confirmRedeemIntent 
[...]

PR #671, opened 2019-03-14
Bounty change unlock period in block height is set to 7-days.

gulshanvasnani:
PR contains following :-
1. `BOUNTY_CHANGE_UNLOCK_PERIOD` constant is set to value which will enable to unlock bounty change period to 7-days.
2. Updated test-cases.
3. Added test-case to verify the unlock period in contract is set to the expected value.

Partially fixes :- #662

### DIFF
--- a/contracts/test/gateway/TestGatewayBase.sol
+++ b/contracts/test/gateway/TestGatewayBase.sol
@@ -31,6 +31,12 @@ import "../../lib/StateRootInterface.sol";
  */
 contract TestGatewayBase is GatewayBase {
 
+    /* Constants */
+
+    /** Unlock period for change bounty in block height. */
+    uint256 public constant BOUNTY_CHANGE_UNLOCK_PERIOD = 100;
+
+
     /* Constructor */
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/test/gateway/TestGatewayBase.sol
```diff
@@ -31,6 +31,12 @@ import "../../lib/StateRootInterface.sol";
  */
 contract TestGatewayBase is GatewayBase {
 
+    /* Constants */
+
+    /** Unlock period for change bounty in block height. */
+    uint256 public constant BOUNTY_CHANGE_UNLOCK_PERIOD = 100;
```
benjaminbollen: this can now be 1 or 2 for the unit test

test/gateway/gateway_base/bounty_change.js
```diff
@@ -205,4 +206,45 @@ contract('GatewayBase.sol', (accounts) => {
       );
     });
   });
+
+  describe('Confirm unlock period', async () => {
+    let gatewayBaseInstance;
+    const owner = accounts[2];
+    const worker = accounts[3];
+    let unlockHeight;
```
benjaminbollen: do we need these initialisations in this `describe`?

Issue #702, opened 2019-04-01
UtilityToken should expect an address without an interface


schemar:
https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/UtilityToken.sol#L79

PR #710, opened 2019-04-11
Contracts: change references from (branded) token to value token

### LINKED ISSUE #701: token should be called valueToken where applicable
Use valueToken instead of token accross contracts (OSTPrime, UtilityToken, EIP20CoGateway, EIP20Gateway and SimpleStake).

Example: https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/UtilityToken.sol#L79

jasonklein:
This PR changes certain references from (branded) token to value token in contracts and tests. #701 says token to value token in contracts, but upon review, it was discovered that there are references to branded token in mosaic-contracts, and tests were changed to be consistent.

Given that the specific example in ticket was a change in `UtilityToken` and given that `EIP20CoGateway` refers to value token and not branded token, changing branded token references seemed consistent with and implicitly within the scope of the ticket—reviewers: please indicate if this understanding is incorrect.

:rotating_light: This PR includes interface changes and introduces getters to support previous ABIs

🚨 This PR introduces changes to the `package-lock.json` (due to removing modules and re-`npm install`-ing)—I am not certain if that is OK, since the `.gitignore` actually says that file should not even be committed (presumably it was tracked _before_ it was added to `.gitignore`).

Resolves #701.


schemar:
Should this go to the parking lot to discuss the interface changes?

jasonklein:
> Should this go to the parking lot to discuss the interface changes?

No need, @schemar—we discussed and agreed on a stand-up to add `token` getter functions for backward compatibility.

Or are we you referring to other interface changes that I've overlooked? 🙈

contracts/gateway/EIP20Gateway.sol
```diff
@@ -1028,6 +1031,20 @@ contract EIP20Gateway is GatewayBase {
...
+     * @dev This function supports previous versions of this contract's ABI
+     *      that expect a public function, token, that returns the address
+     *      of the value token.
+     */
+    function token()
+        external
+        view
+        returns (address valueToken_)
+    {
+        return address(valueToken);
```
schemar: Change one of two:

Line 1044:
```solidity
return (address)
```

Line 1046:
```solidity
valueToken_ = address(valueToken);
```

contracts/gateway/SimpleStake.sol
```diff
@@ -141,6 +141,21 @@ contract SimpleStake {
...
+     * @dev This function supports previous versions of this contract's ABI
+     *      that expect a public function, token, that returns the address
+     *      of the value token.
+     */
+    function token()
+        external
+        view
+        returns (address valueToken_)
+    {
+        return address(valueToken);
```
schemar: Same as above (named return value vs explicit `return`)

contracts/gateway/UtilityToken.sol
```diff
@@ -188,6 +188,20 @@ contract UtilityToken is EIP20Token, Organized, UtilityTokenInterface {
...
+     * @dev This function supports previous versions of this contract's ABI
+     *      that expect a public function, token, that returns the address
+     *      of the value token.
+     */
+    function token()
+        external
+        view
+        returns (address valueToken_)
+    {
+        return address(valueToken);
```
schemar: same

test/gateway/eip20_gateway/constructor.js
```diff
@@ -76,12 +76,21 @@ contract('EIP20Gateway.constructor() ', (accounts) => {
...
 
+    const valueTokenAddress = await gateway.valueToken.call();
+
+    assert.equal(
+      valueTokenAddress,
+      mockToken.address,
+      'Invalid valueTokenAddress address from contract.',
+    );
+
+    // token supports previous ABIs
```
schemar: Should probably be a separate test with its own `it`.

test/gateway/simple_stake/constructor.js
```diff
@@ -43,12 +43,19 @@ contract('SimpleStake.constructor()', (accounts) => {
       'Returned value is not a valid address.',
     );
 
-    const eip20Token = await simpleStake.token.call();
+    const valueToken = await simpleStake.valueToken.call();
+    // token supports previous ABIs
+    const token = await simpleStake.token.call();
```
schemar: Also probably a different test that it supports the old ABI.

test/gateway/simple_stake/constructor.js
```diff
@@ -59,15 +66,15 @@ contract('SimpleStake.constructor()', (accounts) => {
...
       SimpleStake.new(zeroAddress, gateway, { from: accounts[0] }),
-      'Token contract address must not be zero.',
+      'Value token contract address must not be zero.',
     );
   });
 
   it('should fail if zero gateway address is passed', async () => {
-    Utils.expectRevert(
-      SimpleStake.new(token.address, zeroAddress, { from: accounts[0] }),
+    await Utils.expectRevert(
```
schemar: 😳 good catch

test/gateway/utility_token/constructor.js
```diff
@@ -85,12 +85,19 @@ contract('UtilityToken.constructor()', (accounts) => {
...
 
-    const tokenAddress = await utilityToken.token.call();
+    const valueToken = await utilityToken.valueToken.call();
+    assert.strictEqual(
+      valueToken,
+      valueTokenAddress,
+      `Token address from contract must be equal to ${valueTokenAddress}.`,
+    );
 
+    // token supports previous ABIs
```
schemar: Again, I would argue that this should be a different test.

jasonklein:
> Very, very nice. 🎖 Much more understandable in my opinion.
> 
> Just tiny remarks in-line.
> 
> Wasn't there a PR somewhere to change the types of the arguments of tokens to `address` instead of `EIP20Interface`? I may mix that up, though.
> 
> On a different note: is there a place where we define base coin, base token, value token, and utility token? And whether we want to differentiate these terms further for origin/auxiliary and/or OST-gateway/token-gateway?

@schemar: thanks!

Re: re-typing address, you are remembering correctly, but that work was merged (#714) and this PR rebased and conflicts resolved, which is why you see the change in this PR. Apologies for any confusion!

Re: definitions—there is no single place in this repository, but there is a ticket for adding a terminology file to the developer-guidelines with "value token" as the initial entry, https://github.com/OpenST/developer-guidelines/issues/22.

contracts/gateway/EIP20Gateway.sol
jasonklein: We follow the 2nd approach in `EIP20Token`, so I will go with that.

test/gateway/eip20_gateway/constructor.js
jasonklein: @schemar: I don't think we typically have separate tests for getters. Or is your concern that this does not fall under the given test caption, "should pass with correct parameters"? Or is your concern because this function exists for a special reason?

Setting aside the caption, I think the passing constructor test is the appropriate location for testing a getter with an invariant value, because construction is the only call through which the value changes.

Thoughts?

schemar:
> Re: re-typing address, you are remembering correctly, but that work was merged (#714) and this PR rebased and conflicts resolved, which is why you see the change in this PR. Apologies for any confusion!

The original comment was about the fact that the `_valueToken` arguments are still of type `EIP20interface` or am I misunderstanding something here?

test/gateway/eip20_gateway/constructor.js
schemar: My concern was that "backwards compatibility" does not fall under "correct parameters". That's why I suggest a different test about backwards compatibility.

jasonklein: Agreed! But I would argue the test caption should be different. E.g., [in a BrandedToken test](https://github.com/OpenST/brandedtoken-contracts/blob/develop/test/branded_token/constructor.js#L103): `it('Successfully sets state variables'`

If the test caption is the only significant reason, then I'd prefer to leave things in their current locations.

schemar: Yes. If the test is only about testing the variable assignment in the constructor, then I agree 👍

abhayks1:
I am thinking to review this as it was in "In Review column". @schemar if you are already reviewing let me know I will unassign myself.

jasonklein:
> > Re: re-typing address, you are remembering correctly, but that work was merged (#714) and this PR rebased and conflicts resolved, which is why you see the change in this PR. Apologies for any confusion!
> 
> The original comment was about the fact that the `_valueToken` arguments are still of type `EIP20interface` or am I misunderstanding something here?

@schemar: possibly? Issue #702 and PR #714 were only about re-typing from `EIP20Interface` to `address` in `UtilityToken`; other instances of that typing were not changed. In `UtilityToken`, this value is only stored, to represent the connection with a value token, but no functions are called on the value token. However, in those other instances, functions available on the interface are called on value token.

schemar:
> > > Re: re-typing address, you are remembering correctly, but that work was merged (#714) and this PR rebased and conflicts resolved, which is why you see the change in this PR. Apologies for any confusion!
> > 
> > 
> > The original comment was about the fact that the `_valueToken` arguments are still of type `EIP20interface` or am I misunderstanding something here?
> 
> @schemar: possibly? Issue #702 and PR #714 were only about re-typing from `EIP20Interface` to `address` in `UtilityToken`; other instances of that typing were not changed. In `UtilityToken`, this value is only stored, to represent the connection with a value token, but no functions are called on the value token. However, in those other instances, functions available on the interface are called on value token.
> 
> ✏️N.B.: confirming this answer revealed that I missed one 🙈; addressed, rebased, and pushed.

That's just me being confused then 🤷‍♂️ sorry.

@abhayks1 I did a review earlier. I am not sure the same  person has to re-review? 🤷‍♂️

Issue #700, opened 2019-04-01
Gateway nonce should start at zero


schemar:
Currently, the first nonce that the gateway expects from a new user is `1`.
Instead, the first nonce should be `0`.

https://github.com/OpenSTFoundation/mosaic-contracts/blob/fa436b06cacf3ea84d72550b7398f94220476648/contracts/gateway/GatewayBase.sol#L686

Regenerate the proof using the tool from #711

PR #715, opened 2019-04-15
Gateway nonce now starts at zero

deepesh-kn:
Gateway nonce now starts at zero, 
- Updated code
- Updated test proof data 
- Updated unit tests

Fixes: #700


schemar:
Reviewing 👁

test/gateway/eip20_cogateway/redeem.js
```diff
@@ -28,7 +28,7 @@ const MockToken = artifacts.require('MockToken');
 
 const { MessageStatusEnum } = messageBus;
 
-const nonce = new BN(1);
+const nonce = new BN(0);
```
schemar: why hardcode it and not get it from the gateway?

test/gateway/eip20_cogateway/confirm_stake_intent.js
```diff
@@ -275,25 +275,6 @@ contract('EIP20CoGateway.confirmStakeIntent() ', (accounts) => {
     );
   });
 
-  it('should fail when staker nonce is already consumed.', async () => {
```
schemar: Why did you delete this test? Shouldn't it still be tested? (e.g. try to use nonce `0` twice in a row)

test/gateway/eip20_gateway/confirm_redeem_intent.js
```diff
@@ -326,24 +326,6 @@ contract('EIP20Gateway.confirmRedeemIntent()', (accounts) => {
     );
   });
 
-  it('should fail when redeemer nonce is already consumed', async () => {
```
schemar: Why did you delete this test? Shouldn't it still be tested? (e.g. try to use nonce 0 twice in a row)

test/gateway/eip20_gateway/stake.js
```diff
@@ -234,7 +234,7 @@ contract('EIP20Gateway.stake() ', (accounts) => {
     await mockToken.approve(gateway.address, stakeAmount, { from: stakerAddress });
     await baseToken.approve(gateway.address, bountyAmount, { from: stakerAddress });
 
-    nonce = new BN(2);
+    nonce = new BN(1);
```
schemar: hardcoded on purpose? Why not "get nonce"?

test/gateway/eip20_gateway/stake.js
```diff
@@ -259,7 +259,7 @@ contract('EIP20Gateway.stake() ', (accounts) => {
...
```
schemar: hardcoded on purpose? Why not "get nonce"?

test/gateway/eip20_gateway/stake.js
deepesh-kn: It is an existing code. 
I am in favor of using hardcoded values, as we know what should be the nonce.
Any change in the contract logic will break the test.

Why do you think that it should get from "get nonce" ?

test/gateway/eip20_gateway/confirm_redeem_intent.js
deepesh-kn: Similarly,
This test is already covered in `should fail to confirm redeem intent if its already confirmed once`.
In fact, this test was invalid as the redeemer nonce always started with 1.

test/gateway/eip20_cogateway/confirm_stake_intent.js
deepesh-kn: This test is already covered in `should fail to confirm stake intent if its already confirmed once`.
In fact, this test was invalid as the staker nonce always started with 1.

test/gateway/eip20_cogateway/redeem.js
deepesh-kn: Why not hardcoded? 

We already know that the nonce will start at 0. 
This will also assert that nonce starts at 0.

schemar: ok

test/gateway/eip20_cogateway/confirm_stake_intent.js
schemar: ok

Issue #698, opened 2019-04-01
Gateways should store roots in ring buffer


schemar:
To have a constant size `O(1)` instead of a linear size `O(n)` depending on messages that will be sent with the gateways, the storage roots in gateway base should be stored in a circular buffer that prunes old storage roots.

In GatewayBase the messagebox offset should be made public, such that the proof generation can query the offset.

Was potentially done in #669

Regenerate the proof using the tool from #711

sarvesh-ost:
`CircularBufferUint.sol` has an implementation of the circular buffer. 
If the `GatewayBase` extends `CircularBufferUint.sol` then _the position of message box_ in the EVM storage will change because of the variables declared in the `CircularBufferUint.sol`

_The position of message box_ i.e. `MESSAGE_OUTBOX_OFFSET` in the outbox is used to generate Merkel proofs. 
 This change looks like an interface change to me. Maybe there is a work around of this, which I am not aware of.

benjaminbollen:
I suggest we use the inherited contract approach.  It is clearer to understand the storage layout, even if it changes the position.

PR #722, opened 2019-04-29
Gateways should store roots in ring buffer

deepesh-kn:
Fixes: #698
Was potentially done in #669, but this required interface change and the variable index of message box would change.

This PR implements the logic without the interface change. This includes the library.
We can evaluate on which logic should be adapted.

This PR includes the code change, unit tests for the library.

- With Library approach, there is no interface change. No efforts for backward compatibility.
- With the contract approach, the interface will change and the JS layer needs additional effort to provide backward compatibility.

If this approach is acceptable by all then I will add a few more test cases.

### DIFF
--- a/contracts/gateway/GatewayBase.sol
+++ b/contracts/gateway/GatewayBase.sol
@@ -27,6 +27,7 @@ import "../lib/OrganizationInterface.sol";
 import "../lib/Organized.sol";
 import "../lib/SafeMath.sol";
 import "../lib/StateRootInterface.sol";
+import "../lib/CircularBufferUintLib.sol";
 
 /**
  *  @title GatewayBase is the base contract for EIP20Gateway and EIP20CoGateway.
@@ -68,6 +69,9 @@ contract GatewayBase is Organized {
 
     /** Position of message bus in the storage. */
     uint8 constant MESSAGE_BOX_OFFSET = 7;
+
+    /** The max number of storage roots to store. */
+    uint256 private constant MAX_STORAGE_ROOTS = 256;
 
     /**
      * Penalty in bounty amount percentage charged to message sender on revert message.
@@ -149,6 +153,8 @@ contract GatewayBase is Organized {
     mapping(address => bytes32) private outboxActiveProcess;
 
 
+    CircularBufferUintLib.BufferUint storageRootBuffer;
+
     /* Constructor */
 
     /**
@@ -180,6 +186,11 @@ contract GatewayBase is Organized {
         messageBox = MessageBus.MessageBox();
         encodedGatewayPath = '';
         remoteGateway = address(0);
+
+        CircularBufferUintLib.setMaxItemLimit(
+            MAX_STORAGE_ROOTS,
+            storageRootBuffer
+        );
     }
 
 
@@ -223,6 +234,12 @@ contract GatewayBase is Organized {
             "Length of RLP parent nodes is 0"
         );
 
+        // Input block height should be valid.
+        require(
+            _blockHeight > CircularBufferUintLib.head(storageRootBuffer),
+            "Given block height is lower or equal to highest anchored state root block height."
+        );
+
         bytes32 stateRoot = stateRootProvider.getStateRoot(_blockHeight);
 
         // State root should be present for the block height
@@ -257,6 +274,8 @@ contract GatewayBase is Organized {
         );
 
         storageRoots[_blockHeight] = storageRoot;
+        uint256 oldestStoredBlockHeight = CircularBufferUintLib.store(_blockHeight, storageRootBuffer);
+        delete storageRoots[oldestStoredBlockHeight];
 
         // wasAlreadyProved is false since Gateway is called for the first time
         // for a block height
--- a/contracts/gateway/GatewayBase.sol
+++ b/contracts/gateway/GatewayBase.sol
@@ -68,15 +68,15 @@ contract GatewayBase is Organized {
     /* Constants */
 
     /** Position of message bus in the storage. */
-    uint8 constant MESSAGE_BOX_OFFSET = 7;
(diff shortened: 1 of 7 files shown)


schemar:
More persons should comment on this PR.

contracts/lib/CircularBufferUintLib.sol
```diff
@@ -0,0 +1,150 @@
...
+         * zero.
+         */
+        uint256[] items;
+
+        /**
+         * The current index in the items array. The index increases up to
+         * `items.length - 1` and then resets to zero in an endless loop. This
+         * means that a new item will always overwrite the oldest item.
+         */
+        uint256 index;
```
jasonklein: If I correctly understand the purpose of applying this pattern, I would have expected `index` to be typed to something smaller. More specifically, I expected it to be`uint8`, but maybe I'm reading too much into the `MAX_STORAGE_ROOTS` in `GateywayBase` 🤔 

Or we want to allow maximum flexibility?

contracts/lib/CircularBufferUintLib.sol
```diff
@@ -0,0 +1,150 @@
...
+    /* Internal functions */
+
+    /**
+     * @notice Set buffer max item limit/
+     *
+     * @param _maxItems The item to store in the circular buffer.
+     * @param _buffer Uint buffer.
+     */
+    function setMaxItemLimit(
+        uint256 _maxItems,
```
jasonklein: Same question as with `BufferUint.index`—should this be something smaller than `uint256`?

contracts/lib/CircularBufferUintLib.sol
```diff
@@ -0,0 +1,150 @@
...
+        _buffer.items.length = _maxItems;
+    }
+
+    /**
+     * @notice Store a new item in the circular buffer.
+     *
+     * @param _item The item to store in the circular buffer.
+     * @param _buffer The circular buffer.
+     *
+     * @return overwrittenItem_ The item that was in the circular buffer's
```
pgev: Returns cost gas and I am not sure do we need this return variable throughout our code. Caller can always explicitly check if the current value before storing if it's needed.

contracts/lib/CircularBufferUintLib.sol
```diff
@@ -0,0 +1,150 @@
...
+    function store(
+        uint256 _item,
+        BufferUint storage _buffer
+    )
+        internal
+        returns(
+            uint256 overwrittenItem_
+        )
+    {
+        require(
```
pgev: I am not really sure do we need to do this check in all places (I would do it in any other language, however, in "our" world it's going to cost us), however, if we would like to keep, I would propose to make it modifier to improve readability and decrease code duplication.

contracts/lib/CircularBufferUintLib.sol
```diff
@@ -0,0 +1,150 @@
...
+     *         end of the array that represents the circular.
+     *
+     * @param _buffer The circular buffer.
+     */
+    function nextIndex(
+        BufferUint storage _buffer
+    )
+        private
+    {
+        _buffer.index++;
```
pgev: Do you think, we should use SafeMath operations for buffer index?
If yes, we should revert the check possibly:

```solidity
if (_buffer.index == _buffer.items.length.sub(1)) {
    _buffer.index = 0;
} else {
    _buffer.index = _buffer.index.add(1);
}
```

contracts/gateway/GatewayBase.sol
```diff
@@ -257,6 +273,8 @@ contract GatewayBase is Organized {
         );
 
         storageRoots[_blockHeight] = storageRoot;
+        uint256 oldestStoredBlockHeight = CircularBufferUintLib.store(_blockHeight, storageRootBuffer);
+        delete storageRoots[oldestStoredBlockHeight];
```
pgev: I see why have you returned the previously stored element from the circular buffer store function.
What do you think about storing both, blockHeight and storageRoot in circular buffer itself. It would simplify code and make it more robust. Thoughts?

contracts/lib/CircularBufferUintLib.sol
deepesh-kn: Agree, `uint256` very big for this.
Since it's constant I agree it can be `uint8` (or `uint16`).

contracts/lib/CircularBufferUintLib.sol
deepesh-kn: Agree, `uint256` very big for this. I think this should be `uint16`

contracts/lib/CircularBufferUintLib.sol
deepesh-kn: 👍

contracts/lib/CircularBufferUintLib.sol
deepesh-kn: Agree, will make this change once the approach is finalized.
This change is needed in the contract approach as well.

contracts/gateway/GatewayBase.sol
deepesh-kn: If we store both blockHeight and storeageRoot in the circular buffer then, it will become very specific implementation. Currently the circular buffer is generic and can be used for any data other than blockHeight and storageRoot.

contracts/gateway/GatewayBase.sol
```diff
@@ -148,6 +152,7 @@ contract GatewayBase is Organized {
      */
     mapping(address => bytes32) private outboxActiveProcess;
 
+    CircularBufferUintLib.BufferUint storageRootBuffer;
```
benjaminbollen: the name `storageRootBuffer` is badly named because it is storing uint, not storage root

deepesh-kn:
As per the discussion, it is decided to go with the contract approach.
Closing this PR, the new PR is raised here https://github.com/OpenST/mosaic-contracts/pull/726

PR #726, opened 2019-05-06
Gateways should store roots in ring buffer (Contract approach)

deepesh-kn:
Fixes: #698
After the discussion on library vs contract approach (2:3), it was decided to go with the contract approach.
https://github.com/OpenST/mosaic-contracts/pull/722

This is done with the contract approach.

- Updated code.
- Updated proof data.
- Updated unit tests.
- Updated integration tests.
- Updated proof generator tool.

This has interface change. Mosaic.js need to provide backward compatibility.


schemar:
Having a look 👁 

@deepesh-kn what are the differences to #669 ? Would help to know, I think.

deepesh-kn:
> Having a look 👁
> 
> @deepesh-kn what are the differences to #669 ? Would help to know, I think.

Code changes are similar except for that uses the constant. 
`uint256 private constant MAX_STORAGE_ROOTS = 256;`

In this PR, the constructor accepts max storage item size in the constructor as a parameter.

deepesh-kn:
Good catch 👍 


> Looks good to me 🐱. However, additional opinions are required.
> 
> Some general questions:
> 
> * Should there not be a test that asserts that older than `max` state roots are deleted by `GatewayBase`?

Added unit test.

> * Are no other updates required for `MockGatewayBase`?

Update the code.

> * In which test cases is the actual gateway base used for storing (I am confused by the amount of mock and test replacements for the various smart contracts of the gateway)?

I too get confused. its in `test/gateway/gateway_base/prove_gateway.js`

contracts/gateway/GatewayBase.sol
```diff
@@ -67,12 +68,12 @@ contract GatewayBase is Organized {
     /* Constants */
 
     /** Position of message bus in the storage. */
-    uint8 constant MESSAGE_BOX_OFFSET = 7;
+    uint8 constant public MESSAGE_BOX_OFFSET = 9;
```
sarvesh-ost: In the file some places it's `public constant` and here it's `constant public`, not sure which is correct.

contracts/gateway/GatewayBase.sol
```diff
@@ -257,6 +262,8 @@ contract GatewayBase is Organized {
         );
 
         storageRoots[_blockHeight] = storageRoot;
+        uint256 oldestStoredBlockHeight = CircularBufferUint.store(_blockHeight);
+        delete storageRoots[oldestStoredBlockHeight];
```
sarvesh-ost: is this required in the constructor? Before this, no storage has been initialized right?

contracts/gateway/GatewayBase.sol
sarvesh-ost: We should also add getters for internal and private mapping?

contracts/test/gateway/MockGatewayBase.sol
```diff
@@ -125,4 +131,23 @@ contract MockGatewayBase is GatewayBase {
...
+    /**
+     *  @notice Get the storage root for the given block height. This is only
+     *          testing
+     *
+     *  @param _blockHeight Block height for which the storage root is to be
+     *                      fetched.
+     *
+     *  @return storageRoot_ Storage root for the given block height.
+     */
+    function getStorageRoot(
```
sarvesh-ost: This should probably be moved to GatewayBase.sol

test_integration/lib/proof_utils.js
```diff
@@ -21,10 +21,12 @@
 const rlp = require('rlp');
 
 // This is the position of message outbox defined in GatewayBase.sol
-const MESSAGE_OUTBOX_OFFSET = '7';
+// @dev, this is in hex value (in string) without 0x prefix.
+const MESSAGE_OUTBOX_OFFSET = '9';
```
sarvesh-ost: Now constants are public, there shouldn't be any constant on JS layer.

contracts/gateway/GatewayBase.sol
deepesh-kn: This will work even if the storageRoot is not available for the given block height.
What do you propose here? What should be initialized?

contracts/test/gateway/MockGatewayBase.sol
deepesh-kn: Let us evaluate and create a ticket for this.

test_integration/lib/proof_utils.js
deepesh-kn: Let us have this as constant. For testing, it will be better to be a hardcoded value instead of getting from the contract. If it changes in the contract this test should break.

contracts/gateway/GatewayBase.sol
sarvesh-ost: I mean this delete statement is extra here, it's not needed. right?

contracts/test/gateway/MockGatewayBase.sol
sarvesh-ost: OK

test_integration/lib/proof_utils.js
sarvesh-ost: With the getter, we want to achieve backward compatibility right? Even if the position will change, it shouldn't impact the flow right?  isn't that the whole point of having explicit getter? 

Integration test in my mind is more like a user journey test, it shouldn't break if the implementation is intended as backward compatible. 
If a test breaks mean you need to act in some way like update some other code, documentation etc. But I believe this is not the case here.

contracts/gateway/GatewayBase.sol
schemar: This is not in the constructor, this is in `proveGateway`. Here, it is needed to restrict the storage usage of the contract.

sarvesh-ost: Oh yes, My bad. I misunderstood.

PR #728, opened 2019-05-07
Use npm and git submodules

### LINKED ISSUE #712: Use git-submodules to include UtilityToken and Organized contracts
**Use git-submodules to include the following contracts** 
- SafeMath.sol
- EIP20Token.sol
- UtilityToken.sol 
- Organized.sol

**Following is the proposal:**

1. Remove the existing implementation of the above contracts.
2. Use git-submodules to include EIP20Token.sol and UtilityToken.sol from `OpenST/utilitytoken-contracts`.
3. Check if UtilityToken imports the dependent Organized contract.	
4. If not then use `git-submodules` to include Organized contract in the project from `OpenST/organization-contracts`.
5. Use git-submodules to include SafeMath contract in the project from < will be updated >

schemar:
Code deduplication by getting dependencies from npm and git submodules.

* `UtilityToken` contracts are now imported as a git submodule.
* `Organization` contracst are now imported as a git submodule inside
  the `UtilityToken` submodule.
* `SafeMath` is now imported as open zeppelin npm package.

Note that inside the utility token repository, `token` has not been
renamed to `value token`, and therefore some tests needed to be updated.
Also, some tests required updating as `SafeMath` does not give an error
message anymore.

`Organiation` contracts cannot be imported as a separate git submodule
dependency, as then the solidity compiler will complain that there are
multiple definitions of the same symbol (e.g. `contract Organization`).

For now it is fine to import organization from inside utility tokens,
but that will lead to an error latest when there is a diamond
dependency:

```
       Organization
        /        \
UtilityToken   SomethingElse
        \        /
       NewRepository
```

In this case, it couldn't be avoided that `Organization` was defined
multiple times.

To fix this, I created follow-up ticket:
https://github.com/OpenST/developer-guidelines/issues/27

Deduplicating EIP20Token is out-of-scope, as it requires changing
UtilityToken contracts (which has an EIP20Token) and all consumers of
UtilityToken, which currently may or may not rely on the EIP20Token from
there.
Simply adding the npm package does not work, as there will be an error
about multiple definitions of the same Symbol (`contract EIP20Token`).

To fix this, I created a follow-up ticket:
https://github.com/OpenST/developer-guidelines/issues/28

Fixes #712


test/gateway/simple_stake/release_to.js
```diff
@@ -116,7 +116,6 @@ contract('SimpleStake.releaseTo()', (accounts) => {
 
     await Utils.expectRevert(
       simpleStake.releaseTo(beneficiary, releasedAmount, { from: gateway }),
-      'Underflow when subtracting.',
```
deepesh-kn: This is strange that the require error messages are not available.
Was checking the release candidate `v2.3.0-rc.1` of `OpenZeppelin/openzeppelin-solidity` and it looks like the error messages will be available.

package.json
```diff
@@ -32,6 +32,7 @@
     "inquirer": "6.2.1",
     "keccak": "1.4.0",
     "mocha": "5.2.0",
+    "openzeppelin-solidity": "2.1.1",
```
deepesh-kn: why `2.1.1`  instead of the latest `2.2.0` ?

deepesh-kn:
> @deepesh-kn what do you think?

The same version across all repositories sounds good to me 👍

PR #730, opened 2019-05-10
Stake & Mint integration tests with proof

deepesh-kn:
fixes: #607 
This PR Includes the integration test for stake and mint with proof.

### DIFF
--- a/test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
+++ b/test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
@@ -62,7 +62,7 @@ const MessageStatus = {
 };
 
 describe('Stake and mint (with proof)', async () => {
-  before_each(async () => {
+  beforeEach(async () => {
     originWeb3 = shared.origin.web3;
     auxiliaryWeb3 = shared.auxiliary.web3;
     originAccounts = shared.origin.accounts;
@@ -115,7 +115,6 @@ describe('Stake and mint (with proof)', async () => {
     await progressMintWithProof(MessageStatus.Declared);
     await progressStakeWithProof(MessageStatus.Progressed);
   });
-
 });
 
 /**
@@ -137,14 +136,14 @@ const proveGateway = async () => {
   );
 
   // Prove gateway.
-  let tx = await cogateway.proveGateway(
+  const tx = await cogateway.proveGateway(
     new BN(blockNumber),
     outboxProof.encodedAccountValue,
     outboxProof.serializedAccountProof,
     { from: auxiliaryAccounts[0] },
   );
 
-  let event = EventDecoder.getEvents(tx, cogateway);
+  const event = EventDecoder.getEvents(tx, cogateway);
   ProveGatewayAssertion.verify(
     event,
     new BN(blockNumber),
@@ -152,7 +151,7 @@ const proveGateway = async () => {
     gateway.address,
   );
 
-  return {outboxProof, blockNumber: new BN(blockNumber)};
+  return { outboxProof, blockNumber: new BN(blockNumber) };
 };
 
 /**
@@ -173,14 +172,14 @@ const proveCoGateway = async () => {
   );
 
   // Prove gateway.
-  let proofTx = await gateway.proveGateway(
+  const proofTx = await gateway.proveGateway(
     new BN(blockNumber),
     inboxProof.encodedAccountValue,
     inboxProof.serializedAccountProof,
     { from: originAccounts[0] },
   );
 
-  let proofEvent = EventDecoder.getEvents(proofTx, gateway);
+  const proofEvent = EventDecoder.getEvents(proofTx, gateway);
   ProveGatewayAssertion.verify(
     proofEvent,
     new BN(blockNumber),
@@ -188,7 +187,7 @@ const proveCoGateway = async () => {
     cogateway.address,
   );
 
-  return {inboxProof,  blockNumber: new BN(blockNumber)};
+  return { inboxProof, blockNumber: new BN(blockNumber) };
 };
 
 /**
@@ -224,10 +223,10 @@ const stake = async () => {
  * @return {Promise<void>}
  */
 const confirmStakeIntent = async () => {
-  const {outboxProof, blockNumber} = await proveGateway();
+  const { outboxProof, blockNumber } = await proveGateway();
 
   stakeRequest.blockHeight = new BN(blockNumber);
-  let tx = await cogateway.confirmStakeIntent(
+  const tx = await cogateway.confirmStakeIntent(
     stakeRequest.staker,
     stakeRequest.nonce,
     stakeRequest.beneficiary,
@@ -240,7 +239,7 @@ const confirmStakeIntent = async () => {
     { from: auxiliaryAccounts[0] },
   );
 
-  let event = EventDecoder.getEvents(tx, cogateway);
+  const event = EventDecoder.getEvents(tx, cogateway);
   // Assert event.
   ConfirmStakeIntentAssertion.verify(event, stakeRequest);
 };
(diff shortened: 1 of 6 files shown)


test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
```diff
@@ -0,0 +1,330 @@
...
+const MessageStatus = {
+  Undeclared: 0,
+  Declared: 1,
+  Progressed: 2,
+  DeclaredRevocation: 3,
+  Revoked: 4,
+};
+
+describe('Stake and mint (with proof)', async () => {
+  before_each(async () => {
```
sarvesh-ost: Does this work 😱 or its `beforeEach` 🤔

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
```diff
@@ -0,0 +1,330 @@
...
+  );
+
+  const event = EventDecoder.getEvents(tx, gateway);
+  // Assert event and balances.
+  await assertProgressStake.verify(event, stakeRequest, initialBalancesBeforeProgress, true);
+};
+
+/**
+ * Progress mint with proof.
+ * @param {number} gatewayInboxMessageStatus Outbox message status in gateway.
```
sarvesh-ost: Documentation is misaligned with the variable

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
```diff
@@ -0,0 +1,330 @@
...
+ * @return {Promise<void>}
+ */
+const progressMintWithProof = async (gatewayOutboxMessageStatus) => {
+  const {outboxProof, blockNumber} = await proveGateway();
```
sarvesh-ost:
```suggestion
  const { outboxProof, blockNumber } = await proveGateway();
```

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
```diff
@@ -0,0 +1,330 @@
+// Copyright 2019 OpenST Ltd.
```
sarvesh-ost: Can you fix linting issues of this file.

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
deepesh-kn: it works :) 
Changed to `beforeEach`

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
deepesh-kn: Fixed.

test_integration/02_stake_and_mint/utils/progress_mint_assertion.js
```diff
@@ -80,9 +81,10 @@ class ProgressMintAssertion {
      * @param {StakeRequest} stakeRequest Stake request.
      * @param {balance} initialBalances Initial baseToken and token balances of
      *                                  beneficiary, ostPrime and coGateway.
+     * @param {Boolean} proofProgress `true` if progress is with with proof.
```
schemar:
```suggestion
     * @param {Boolean} proofProgress `true` if progress is with proof.
```

schemar:
Reviewing 🐴

test_integration/02_stake_and_mint/utils/progress_mint_assertion.js
```diff
@@ -218,14 +221,22 @@ class ProgressMintAssertion {
...
     );
-    assert.strictEqual(
-      eventData._unlockSecret,
-      stakeRequest.unlockSecret,
-      'Unlock secret must match.',
-    );
+    if (proofProgress === true) {
+      assert.strictEqual(
+        eventData._unlockSecret,
+        Utils.ZERO_BYTES32,
```
schemar: How do you know here that the secret was zero bytes? Doesn't that depend on where this method was called from? Should probably be a function argument here. What do you think?

Or rather: why is the unlock secret in the event and zero when proof is progressed?

test_integration/02_stake_and_mint/utils/progress_mint_assertion.js
```diff
@@ -218,14 +221,22 @@ class ProgressMintAssertion {
...
-      false,
+      proofProgress,
       'Proof progress flag should be false.',
     );
-    assert.strictEqual(
-      eventData._unlockSecret,
-      stakeRequest.unlockSecret,
-      'Unlock secret must match.',
-    );
+    if (proofProgress === true) {
```
schemar: Didn't you assert in the lines above that it was false? 😕 I think the message in the assert above is wrong.

test_integration/02_stake_and_mint/utils/progress_stake_assertion.js
```diff
@@ -208,14 +212,22 @@ class ProgressStakeAssertion {
...
-      eventData._unlockSecret,
-      stakeRequest.unlockSecret,
-      'Unlock secret must match.',
+      proofProgress,
+      `Proof progress flag should be ${proofProgress}.`,
     );
+    if (proofProgress === true) {
+      assert.strictEqual(
+        eventData._unlockSecret,
+        Utils.ZERO_BYTES32,
```
schemar: Same as above. Maybe a comment could explain it?

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
```diff
@@ -0,0 +1,329 @@
...
+    await confirmStakeIntent();
+    await progressStakeWithProof(MessageStatus.Declared);
+    await progressMintWithProof(MessageStatus.Progressed);
+  });
+
+  it('Stake & Mint (progress mint and then progress stake)', async () => {
+    await stake();
+    await confirmStakeIntent();
+    await progressMintWithProof(MessageStatus.Declared);
+    await progressStakeWithProof(MessageStatus.Progressed);
```
schemar: Progressing mint before stake (with proof)? 😕

test_integration/02_stake_and_mint/utils/progress_mint_assertion.js
deepesh-kn: > How do you know here that the secret was zero bytes? Doesn't that depend on where this > > method was called from? Should probably be a function argument here. What do you think?

`proofProgress` is already a param in the function argument, its value depends on where this method is called from.  it is used to assert different values. 

> Or rather: why is the unlock secret in the event and zero when proof is progressed?

In this way, the facilitator will subscribe to a single event.
If the `progressXXXwithProof` is called, in that case, the Merkle proof is used instead of `unlockSecret`. The event name and params emitted are the same in both `progressXXX` and `progressXXXwithProof` function, so the unlockSecret value will be bytes32(0) as there is no unlock secret. Facilitators can read `_proofProgress` param value to determine if the progress was with the unlock secret or proof data.

test_integration/02_stake_and_mint/utils/progress_mint_assertion.js
deepesh-kn: Thanks 👍 
I corrected the assert messages for both asserts.

test_integration/02_stake_and_mint/utils/progress_stake_assertion.js
deepesh-kn: Added

test_integration/02_stake_and_mint/02_stake_and_mint_proof.js
deepesh-kn: That is possible.

schemar: I thought you need to prove that stake was progressed? I am confused ☹️

PR #745, opened 2019-05-17
Staker proxy

### LINKED ISSUE #741: Staker proxy contract.
This is closely dependent on #740 

**StakerProxy contract:**

**Please note:** the contract is referred to `StakerProxy`, but this is not final name. Suggest a better name.

Gateways can do one stake - mint process at a time for a given staker address. So the composer contract cannot call gateway.stake() function, it will be not able to serve all stake requests from multiple stakers addresses. So a proxy contract is introduced as a proposal here.

- This contract has a minimal code, that gets deployed from the `Composer` contract.
- It is deployed per user once.
- It can be used for arbitrary gateways.
- Actual staker is the owner, and it can be destroyed by the owner only.
- All other functions can be called only by Composer contract.

_______________________________
This contract can have following variables.
```
address  public composer;
address  public owner;
```
Constructor
```  
    constructor(
        address _composer, 
        address _owner
    ) 
        public;
```
Accept stake request
```
     function acceptStakeRequest(
        uint256 _amount,
        address _beneficiary,
        uint256 _gasPrice,
        uint256 _gasLimit,
        uint256 _nonce,
        bytes32 _hashLock,
        address _staker,
        address _gateway
    )
        onlyComposer
        external
        returns (bytes32 messageHash_);
```

More detail is on EPIC

schemar:
A new staker proxy contract to stake on the gateway.

Temporarily added a composer interface to be able to call `removeStakerProxy`.
See also comment on https://github.com/OpenST/mosaic-contracts/issues/742.

Added unit tests for the proxy's methods.
Added new test doubles to ease writing of unit tests.

Fixes #741

### DIFF
--- a/contracts/gateway/EIP20GatewayInterface.sol
+++ b/contracts/gateway/EIP20GatewayInterface.sol
@@ -57,7 +57,7 @@ interface EIP20GatewayInterface {
     /**
      * @notice Get the nonce for the given account address.
      *
-     * @param _account Account address for which the nonce is to fetched.
+     * @param _account Account address for which the nonce is to be fetched.
      *
      * @return The nonce.
      */
--- a/contracts/gateway/EIP20GatewayInterface.sol
+++ b/contracts/gateway/EIP20GatewayInterface.sol
@@ -1,0 +1,89 @@
+pragma solidity ^0.5.0;
+
+// Copyright 2019 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+import "../lib/EIP20Interface.sol";
+
+
+interface EIP20GatewayInterface {
+
+    /**
+     * @notice Initiates the stake process. In order to stake the staker
+     *         needs to approve Gateway contract for stake amount.
+     *         Staked amount is transferred from staker address to
+     *         Gateway contract. Bounty amount is also transferred from staker.
+     *
+     * @param _amount Stake amount that will be transferred from the staker
+     *                account.
+     * @param _beneficiary The address in the auxiliary chain where the utility
+     *                     tokens will be minted.
+     * @param _gasPrice Gas price that staker is ready to pay to get the stake
+     *                  and mint process done.
+     * @param _gasLimit Gas limit that staker is ready to pay.
+     * @param _nonce Nonce of the staker address.
+     * @param _hashLock Hash Lock provided by the facilitator.
+     *
+     * @return messageHash_ Message hash is unique for each request.
+     */
+    function stake(
+        uint256 _amount,
+        address _beneficiary,
+        uint256 _gasPrice,
+        uint256 _gasLimit,
+        uint256 _nonce,
+        bytes32 _hashLock
+    )
+        external
+        returns (bytes32 messageHash_);
+
+    /**
+     * @notice Get the nonce for the given account address.
+     *
+     * @param _account Account address for which the nonce is to fetched.
+     *
+     * @return The nonce.
+     */
+    function getNonce(address _account)
+        external
+        view
+        returns (uint256);
+
+    /**
+     * @notice Amount of EIP20 which is staked by facilitator.
+     *
+     * @return The amount.
+     */
+    function bounty() external view returns (uint256);
+
+    /**
+     * @notice Get the value token of this gateway.
+     *
+     * @return The address of the value token.
+     */
+    function valueToken() external view returns (EIP20Interface);
+
+    /**
+     * @notice Get the base token of this gateway.
+     *
+     * @return The address of the base token.
+     */
+    function baseToken() external view returns (EIP20Interface);
+}
--- a/contracts/gateway/StakerProxy.sol
+++ b/contracts/gateway/StakerProxy.sol
@@ -1,0 +1,200 @@
+pragma solidity ^0.5.0;
+
+// Copyright 2019 OpenST Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+import "./ComposerInterface.sol";
+import "./EIP20Gateway.sol";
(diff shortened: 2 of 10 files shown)


contracts/gateway/StakerProxy.sol
```diff
@@ -0,0 +1,200 @@
...
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+import "./ComposerInterface.sol";
+import "./EIP20Gateway.sol";
```
deepesh-kn: Can this be an interface instead?

schemar: OK I will create a new gateway that will be implemented by the gateway.

schemar:
@deepesh-kn I added a gateway, please check.
I will now start with the tests.

schemar:
Roughly 30 seconds faster 😐

contracts/gateway/StakerProxy.sol
```diff
@@ -0,0 +1,200 @@
...
+        uint256 _nonce,
+        bytes32 _hashLock,
+        EIP20GatewayInterface _gateway
+    )
+        external
+        onlyComposer
+        mutexed
+        returns (bytes32 messageHash_)
+    {
+        approveTransfers(_gateway, _amount);
```
deepesh-kn: We can do `approveTransfers` after the nonce validations

contracts/gateway/EIP20GatewayInterface.sol
```diff
@@ -0,0 +1,89 @@
...
+     * @return The amount.
+     */
+    function bounty() external view returns (uint256);
+
+    /**
+     * @notice Get the value token of this gateway.
+     *
+     * @return The address of the value token.
+     */
+    function valueToken() external view returns (EIP20Interface);
```
deepesh-kn: Should we return `address`? 
The caller can type cast it to EIP20Interface. What do you think?

deepesh-kn:
> Roughly 30 seconds faster 😐

? 🤔

contracts/gateway/EIP20GatewayInterface.sol
schemar: It requires the caller to know implementation details about the gateway behind the interface. I would say the interface must suffice to use it.

contracts/gateway/StakerProxy.sol
schemar: Will move.

contracts/gateway/EIP20GatewayInterface.sol
schemar: Also, and more importantly, you must be able to replace different implementations of the same interface with one another. In this case, this is only possible if it returns an EIP20 interface here.

schemar: We could create a separate interface that only has the methods that we require here, e.g. "approve".

schemar:
Ready for review

schemar:
🎈bouncing due to failing tests

schemar:
Back to 🔍Review, only a ganache hiccup.

abhayks1:
Reviewing...

contracts/gateway/StakerProxy.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * @notice Transfers EIP20 token to destination address.
+     *
+     * @param _token EIP20 token address.
+     * @param _to Address to which tokens are transferred.
+     * @param _value Amount of tokens to be transferred.
+     */
+    function transferToken(
+        EIP20Interface _token,
```
abhayks1: Since token can be any EIP20 token, require can be added that token should not be address 0.
Also please add respective unit case.

test/gateway/staker_proxy/transfer_token.js
```diff
@@ -0,0 +1,90 @@
...
+      stakerProxy.transferToken(
+        someoneElse,
+        someoneElse,
+        transferValue,
+        { from: someoneElse },
+      ),
+      'This function can only be called by the owner.',
+    );
+  });
+
```
abhayks1: Negative unit case can be added for use case "when owner doesn't have sufficient balance and he is doing transferToken".

test/gateway/staker_proxy/stake.js
```diff
@@ -0,0 +1,191 @@
...
+      nonce,
+      hashLock,
+      spyGateway.address,
+      { from: composer },
+    );
+
+    assert.strictEqual(
+      messageHash,
+      (await spyGateway.messageHash.call()),
+      'The spy did not record the correct amount staked.',
```
abhayks1: Is this error message correct 😕 ?

test/gateway/staker_proxy/stake.js
```diff
@@ -0,0 +1,191 @@
...
+    assert.strictEqual(
+      (await spyGateway.beneficiary.call()),
+      beneficiary,
+      'The spy did not record the correct beneficiary.',
+    );
+
+    assert.strictEqual(
+      (await spyGateway.gasPrice.call()).toString(10),
+      gasPrice,
+      'The spy did not record the correct gas price staked',
```
abhayks1: Full stop can be added in the end of error messages 👁 . Similarly for others messages .

test/gateway/staker_proxy/stake.js
```diff
@@ -0,0 +1,191 @@
...
+      'The spy did not record the correct approval to',
+    );
+
+    assert.strictEqual(
+      (await spyValueToken.approveAmount.call()).toString(10),
+      amount,
+      'The spy did not record the correct approval amount',
+    );
+
+
```
abhayks1: Two new lines are not needed here.

schemar:
On it 🚂

test/gateway/staker_proxy/transfer_token.js
schemar: In this case the unit under test is the `StakerProxy`. It is not in the scope of the `StakerProxy` to test the token that is used for the transfer (and should fail on insufficient funds).

What do you think? Should I still add a test? I wonder which revert message is returned, the one from the token or the one from the surrounding `StakerProxy` 🤔

contracts/gateway/EIP20GatewayInterface.sol
schemar: @deepesh-kn resolved or further points to discuss?

deepesh-kn:
Reviewing 🦖

contracts/gateway/EIP20GatewayInterface.sol
deepesh-kn: We can keep your implementation. 👍

contracts/gateway/StakerProxy.sol
schemar: It is about the `_token` being `address(0).

deepesh-kn: > It is about the _token being address(0). Good question about _to, though 🤔

Yes, we can never be sure if the passed address is of valid EIP20 contract, nor is `address(0)`. 
We had this check in `GatewayComposer`, so we can retain the check for `address(0)`. 

`_to` can be `address(0)`

schemar: ok I will add the redundant `require` 👍

schemar:
Ready for review again. `_to` can now be zero, `_token` cannot.

test/gateway/staker_proxy/transfer_token.js
abhayks1: Error will be thrown from token transfer method because of SafeMath -ve calculation.
https://github.com/OpenST/mosaic-contracts/blob/develop/contracts/gateway/EIP20Token.sol#L257
I think we will get "VM exception" error 🤔 .

No need to add this test as it's out of scope of StakerProxy.

gulshanvasnani:
@schemar , The logic which is used for implementing mutex already exists in [Mutex](https://github.com/OpenST/mosaic-contracts/blob/develop/contracts/lib/Mutex.sol) contract under lib. I think we must use the contract.

schemar:
> @schemar , The logic which is used for implementing mutex already exists in [Mutex](https://github.com/OpenST/mosaic-contracts/blob/develop/contracts/lib/Mutex.sol) contract under lib. I think we must use the contract.

Good catch, thanks 👍 Can you please create the appropriate ticket?

PR #746, opened 2019-05-17
Parallelize Travis CI

pgev:


schemar:
Contracts are also compiled multiple times on each run. Maybe that can be optimized in a way.

pgev:
>Contracts are also compiled multiple times on each run. Maybe that can be optimized in a way.

@schemar Definitely. I would propose to have a warm up job and then run tests.

schemar:
Is that possible? To share state of a warmup job with subsequent parallel test VMs?

schemar:
I suggest we merge this as is. It already speeds up the tests to take only half the time.

PR #750, opened 2019-05-21
Composer contract

### LINKED ISSUE #740: OST' Composer contract
I am not sure if this ticket needs to be broken into the smaller ticket.

**Composer contract:**
**Please note:** the contract is referred to `Composer`, but this is not final name. Suggest a better name.

- `Composer` contract will be `Organized`
- Single `Composer` contract must work for arbitrary gateway.
- Any address(user) must be able to request stake for OST on origin chain by calling `requestStake` function.
```
function requestStake(
        address _gateway,
        uint256 _amount,
        address _beneficiary,
        uint256 _gasPrice,
        uint256 _gasLimit,
        uint256 _nonce
    )
        external
        returns (
            bytes32 stakeRequestHash_
        )
```
- Composer should have a function that returns the current `nonce` for the given address that can be used for `stakeRequest`
```
function getNonce(
        address _account,
        address _gateway
    ) 
        external 
        view
        returns (uint256 nonce_)
```

- Facilitators (whitelisted, for now, later it can be open) can facilitate the stake and mint process on behalf of the user by providing the hash lock. A whitelisted address can call `acceptStakeRequest` to facilitate stake and mint.
```
function acceptStakeRequest(
        bytes32 _stakeRequestHash,
        bytes32 _hashLock
    )
        onlyWorker // only whitelisted address can call this
        external
        returns (bytes32 messageHash_)
```
- The staker can revoke the stake request any time unless it has been accepted and processed by the facilitator. Staker can call `revokeStakeRequest()` function to revoke the stake request. Staker can revoke only its own stake requests.
```
function revokeStakeRequest(
        bytes32 _stakeRequestHash
    )
        external 
```
- A facilitator can reject the stakeRequest by calling `rejectStakeRequest`. We should evaluate if this function is needed. If the facilitator pool will be open then we may not need this function.
```
function rejectStakeRequest(
        bytes32 _stakeRequestHash
    )
        onlyWorker // only whitelisted address can call this
        external
```
- All events of the composer contracts must be sufficient for the facilitator to decide if they want to accept or ignore the facilitation.
```
<To be added>
```
- Currently, there is a limitation of one stake request at a time in the Gateway contracts. This must be gracefully handled in the composer contract. One of the proposals is given below that uses proxy contract for staker.
- The organization can deactivate the new stake requests. By doing this no one can call stakeRequest. Facilitators can still accept stake requests. Staker can still revoke the stake requests.

gulshanvasnani:
PR contains :
1.  OSTComposer contract which facilitates the staker to get OSTPrime on the sidechain.
2. Unit test cases for `requestStakeRequest`, `acceptStakeRequest`, `revokeStakeRequest`, `rejectStakeRequest` and `removeStakerProxy` methods.

Fixes #740


contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        external
+        onlyWorker
+        returns(bool success_)
+    {
+        StakeRequest storage stakeRequest = stakeRequests[_stakeRequestHash];
+        require(stakeRequest.staker == address(0),"Invalid stake request hash.");
+
+        address staker = stakeRequests[_stakeRequestHash].staker;
+        activeStakeRequestCount[staker].sub(1);
+        address gateway = stakeRequests[_stakeRequestHash].gateway;
```
schemar:
```suggestion
        EIP20GatewayInterface gateway = stakeRequests[_stakeRequestHash].gateway;
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+
+        delete stakeRequests[_stakeRequestHash];
+        delete stakeRequestHashes[stakeRequest.staker][stakeRequest.gateway];
+
+        EIP20Interface token = EIP20GatewayInterface(gateway).valueToken();
```
schemar:
```suggestion
        EIP20Interface valueToken = gateway.valueToken();
```

schemar: If you get the token before the deletes together with the gateway, it make more sense to the reader. Right now the gateway somewhat dangles mid-air up there.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+
+        uint256 amount = stakeRequests[_stakeRequestHash].amount;
```
schemar: You already deleted the stake request from the mapping at this point!!!

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        uint256 _gasPrice,
+        uint256 _gasLimit,
+        uint256 _nonce
+    )
+        private
+        pure
+        returns(bytes32 stakeRequestIntentHash_)
+    {
+        stakeRequestIntentHash_ = keccak256(abi.encodePacked(
+            STAKEREQUEST_INTENT_TYPEHASH,
```
schemar: I think the domain hash is missing, please re-check the EIP.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        uint256 nonce;
+        address staker;
+        address gateway;
+    }
+
+
+    /* Public Variables */
+
+    /* Stores hash of the stakerequest per staker to gateway. */
+    mapping (address => mapping(address => bytes32))  public stakeRequestHashes;
```
schemar: How does this even happen??
```suggestion
    mapping (address => mapping(address => bytes32)) public stakeRequestHashes;
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+            beneficiary: _beneficiary,
+            gasPrice: _gasPrice,
+            gasLimit: _gasLimit,
+            nonce: _nonce,
+            staker: msg.sender,
+            gateway: _gateway
+        });
+        activeStakeRequestCount[msg.sender].add(1);
+
+        EIP20Interface token = EIP20GatewayInterface(_gateway).valueToken();
```
schemar:
```suggestion
        EIP20Interface valueToken = EIP20GatewayInterface(_gateway).valueToken();
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+    {
+        require(
+            _stakeRequestHash == bytes32(0),
+            "Stake request hash is invalid"
+        );
+
+        StakeRequest storage stakeRequest = stakeRequests[_stakeRequestHash];
+        require(
+            stakeRequest.staker != address(0),
+            "Stake request must exists."
```
schemar: This check makes the last `require` redundant.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        );
+
+        EIP20GatewayInterface gateway = EIP20GatewayInterface(stakeRequest.gateway);
+
+        StakerProxy stakerProxy = stakerProxies[stakeRequest.staker];
+
+        EIP20Interface token = gateway.valueToken();
```
schemar:
```suggestion
        EIP20Interface valueToken = gateway.valueToken();
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        require(
+            token.transfer(address(stakerProxy), stakeRequest.amount),
+            "Staked amount must be transferred to the staker proxy."
+        );
+
+        EIP20Interface baseToken = EIP20GatewayInterface(gateway).baseToken();
```
schemar:
```suggestion
        EIP20Interface baseToken = gateway.baseToken();
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        uint256 bounty = gateway.bounty();
+        require(
+            baseToken.transferFrom(msg.sender, address(stakerProxy), bounty),
+            "Bounty amount must be transferred to stakerProxy."
+        );
+
+        messageHash_ = StakerProxy(stakerProxy).stake(
```
schemar:
```suggestion
        messageHash_ = stakerProxy.stake(
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        external
+        returns(bool success_)
+    {
+        address staker = stakeRequests[_stakeRequestHash].staker;
+        require(
+            staker == address(0),
+            "Invalid stake request hash."
+        );
+        require(
+            staker == msg.sender,
```
schemar: These two requires make absolutely no sense in their combination. I think you want to delete the first one?

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+
+        activeStakeRequestCount[msg.sender].sub(1);
+
+        address gateway = stakeRequests[_stakeRequestHash].gateway;
+        delete stakeRequestHashes[staker][gateway];
+        delete stakeRequests[_stakeRequestHash];
+
+        EIP20Interface token = EIP20GatewayInterface(gateway).valueToken();
+
+        uint256 amount = stakeRequests[_stakeRequestHash].amount;
```
schemar: You already deleted the stake request above!!!

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+            "Only valid staker can revoke the stake request."
+        );
+
+        activeStakeRequestCount[msg.sender].sub(1);
+
+        address gateway = stakeRequests[_stakeRequestHash].gateway;
+        delete stakeRequestHashes[staker][gateway];
+        delete stakeRequests[_stakeRequestHash];
+
+        EIP20Interface token = EIP20GatewayInterface(gateway).valueToken();
```
schemar: If you get the token before the deletes together with the gateway, it make more sense to the reader. Right now the gateway somewhat dangles mid-air up there.

schemar:
```suggestion
        EIP20Interface valueToken = EIP20GatewayInterface(gateway).valueToken();
```

contracts/gateway/OSTComposer.sol
gulshanvasnani: As the hashes are not signed/verified, we need not add domain hash.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+        uint256 gasLimit;
+        uint256 nonce;
+        address staker;
+        address gateway;
+    }
+
+
+    /* Public Variables */
+
+    /* Stores hash of the stakerequest per staker to gateway. */
```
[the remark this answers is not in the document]
gulshanvasnani: It is the mapping of staker to gateway address for storing stakeRequestHash.
I have updated the docs.

deepesh-kn:
Reviewing :🦖

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+    struct StakeRequest {
+        uint256 amount;
+        address beneficiary;
+        uint256 gasPrice;
+        uint256 gasLimit;
+        uint256 nonce;
+        address staker;
+        address gateway;
+    }
+
```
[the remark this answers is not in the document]
gulshanvasnani: Added the documentation to events and structs.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+            "Invalid stake request hash."
+        );
+        require(
+            staker == msg.sender,
+            "Only valid staker can revoke the stake request."
+        );
+
+        activeStakeRequestCount[msg.sender].sub(1);
+
+        address gateway = stakeRequests[_stakeRequestHash].gateway;
```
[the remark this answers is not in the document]
gulshanvasnani: Changed the type of `gateway` in `StakeRequest` struct to `EIP20GatewayInterface`.

contracts/test/gateway/SpyToken.sol
```diff
@@ -32,11 +32,46 @@ contract SpyToken {
...
         approveFrom = msg.sender;
         approveTo = _to;
         approveAmount = _amount;
 
         success_ = true;
     }
-}
+
+    function transferFrom(address _from, address _to, uint256 _amount) external returns (bool success_) {
+        if(transferFromFakeResponse) {
```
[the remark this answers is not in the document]
gulshanvasnani: Even this works for me.
Changed it to suggested approach.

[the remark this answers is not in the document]
gulshanvasnani: I think we should not the values for `transferAddress`, `toAddress` and `transferAmount` in case of failure.
What do you think ?

deepesh-kn:
> Generally, I am not convinced that it is beneficial to track and check the nonces in the composer and the staker proxy. What is the benefit of providing and repeatedly checking it? /cc @deepesh-kn

Maybe we don't need at both places.
If we don't check it in `stakerProxy.stake()` then still, it will work because it will fail in `gateway.stake()`

We can keep it in Composer's request stake function. This can ensure that the data stored is correct and we are increasing the storage with correct data only.

What do you think? 
cc: @schemar

test/ost_composer/accept_stake.js
```diff
@@ -0,0 +1,182 @@
...
+    assert.strictEqual(
+      (activeGatewayRequestCountBeforeAS.sub(activeGatewayRequestCountAfterAS)).eqn(1),
+      true,
+      `Expected active gateway request for ${stakeRequest.staker} is `
+      + `${activeGatewayRequestCountBeforeAS.sub(activeGatewayRequestCountAfterAS)}`
+      + `but got ${activeGatewayRequestCountAfterAS}`,
+    );
+
+    // Verifying the storage stakeRequestHash and stakeRequests. After the revokeStakeRequest
+    // is successful, these two storage are references are cleared.
```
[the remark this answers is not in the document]
gulshanvasnani: After the `acceptStakeRequest` is successful, storage stakeRequestHash and stakeRequests is deleted for that particular stakeRequestHash of the staker.
I have corrected the comment.

schemar:
I'm also re-reviewing 🦑

schemar:
> > Generally, I am not convinced that it is beneficial to track and check the nonces in the composer and the staker proxy. What is the benefit of providing and repeatedly checking it? /cc @deepesh-kn
> 
> Maybe we don't need at both places.
> If we don't check it in `stakerProxy.stake()` then still, it will work because it will fail in `gateway.stake()`
> 
> We can keep it in Composer's request stake function. This can ensure that the data stored is correct and we are increasing the storage with correct data only.
> 
> What do you think?
> cc: @schemar

Since only the composer can call `stake` on the staker proxy, this would work. The nonce on the gateway cannot change from somewhere else, so a check on the composer makes sense 👍

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,436 @@
...
+    {
+        StakeRequest storage stakeRequest = stakeRequests[_stakeRequestHash];
+        require(
+            stakeRequest.staker != address(0),
+            "Stake request must exists."
+        );
+
+        EIP20GatewayInterface gateway = stakeRequest.gateway;
+
+        StakerProxy stakerProxy = stakerProxies[address(stakeRequest.staker)];
```
schemar: Doesn't this work? `stakeRequest.staker` is stored as `address` in the Struct.

```suggestion
        StakerProxy stakerProxy = stakerProxies[stakeRequest.staker];
```

deepesh-kn:
> > I think you don't need the storage variable stakeRequestHashes. Instead, you could use the created stake hash to check if a stake request is already stored at that hash. That saves storage and reduces the possibilities for errors.

Interesting point. I think we can take this up in the next sprint.
There are a few more enhancements/optimizations that we can think of, we will create new tickets.

sarvesh-ost:
> There are a few more enhancements/optimizations that we can think of, we will create new tickets.

I didn't understand the part, Why it's not needed? 🤔 

How to ensure that only one stake request is allowed to be opened in the ost composer for a given gateway? 
it's currently ensured by ` stakeRequestHashes[msg.sender][address(_gateway)] == bytes32(0),` on L186

contracts/gateway/OSTComposer.sol
gulshanvasnani: Thanks. It works.
Changed it.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,437 @@
...
+
+        StakerProxy stakerProxy = stakerProxies[msg.sender];
+
+        if(address(stakerProxy) == address(0)) {
+            stakerProxy = new StakerProxy(msg.sender);
+            stakerProxies[msg.sender] = stakerProxy;
+        }
+
+        require(
+            _gateway.getNonce(address(stakerProxy)) == _nonce,
```
sarvesh-ost: `Nonce` is being checked here as well as staker proxy. It seems like duplicate logic. 
cc/ @schemar

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,437 @@
...
+     * @param _beneficiary The address in the auxiliary chain where the utility
+     *                     tokens will be minted.
+     * @param _gasPrice Gas price that staker is ready to pay to get the stake
+     *                  and mint process done.
+     * @param _gasLimit Gas limit that staker is ready to pay.
+     * @param _nonce Staker nonce specific to gateway.
+     *
+     * @return stakeRequestHash_ A unique hash for stake request.
+     */
+    function requestStake(
```
sarvesh-ost: This method should probably check that the gateway doesn't have any open stake request for the staker proxy. Otherwise, accept could fail.

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+
+    /* Public Variables */
+
+    /* Mapping of staker to gateway to store stake request hash. */
+    mapping (address => mapping(address => bytes32)) public stakeRequestHashes;
+
+    /* Mapping of staker addresses to their StakerProxy. */
+    mapping (address => StakerProxy) public stakerProxies;
```
sarvesh-ost: I think both mappings can be combined to a single object as both are details about a staker. 
```
mapping (address => StakerProxy) public stakerProxies;
mapping(address => uint256) public activeStakeRequestCount;
```
Also They are accessed together in `removeStakerProxy`.

contracts/gateway/OSTComposer.sol
gulshanvasnani: Ticket [OSTComposer optimization](https://github.com/OpenST/mosaic-chains/issues/79) to address above is created.

contracts/gateway/OSTComposer.sol
gulshanvasnani: Same as above.

schemar:
> > There are a few more enhancements/optimizations that we can think of, we will create new tickets.
> 
> I didn't understand the part, Why it's not needed? 🤔
> 
> How to ensure that only one stake request is allowed to be opened in the ost composer for a given gateway(before accept stake)?
> it's currently ensured by ` stakeRequestHashes[msg.sender][address(_gateway)] == bytes32(0),` on L186

You can assert the same like so (after building the hash):

```solidity
    stakeRequestHash_ = hashStakeRequest(
        msg.sender,
        address(_gateway),
        _amount,
        _beneficiary,
        _gasPrice,
        _gasLimit,
        _nonce
    );

    require(
        stakeRequests[stakeRequestHash_].staker == address(0),
        "Request for this staker at this gateway is already in process."
    );
```

contracts/gateway/OSTComposer.sol
```diff
@@ -0,0 +1,427 @@
...
+     *
+     * @param _stakeRequestHash Hash of the stake request.
+     *
+     * @return success_ `true` if stake request is deleted.
+     */
+    function rejectStakeRequest(
+        bytes32 _stakeRequestHash
+    )
+        external
+        onlyWorker
```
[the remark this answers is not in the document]
deepesh-kn: > This means the facilitators have to trust each other when they are registered with the same organization. Is that ok? Just a thought.
> 

Currently yes, maybe `rejectStakeRequest` is not needed at all.

sarvesh-ost:
> ```solidity
> stakeRequestHash_
> ```

Stake request hash will be different if any parameter will change like `amount`, `gasPrice`,  `gasLimit` in the next request.

sarvesh-ost:
> > > There are a few more enhancements/optimizations that we can think of, we will create new tickets.
> > 
> > 
> > I didn't understand the part, Why it's not needed? 🤔
> > How to ensure that only one stake request is allowed to be opened in the ost composer for a given gateway(before accept stake)?
> > it's currently ensured by ` stakeRequestHashes[msg.sender][address(_gateway)] == bytes32(0),` on L186
> 
> You can assert the same like so (after building the hash):
> 
> ```solidity
>     stakeRequestHash_ = hashStakeRequest(
>         msg.sender,
>         address(_gateway),
>         _amount,
>         _beneficiary,
>         _gasPrice,
>         _gasLimit,
>         _nonce
>     );
> 
>     require(
>         stakeRequests[stakeRequestHash_].staker == address(0),
>         "Request for this staker at this gateway is already in process."
>     );
> ```

Stake request hash will be different if any parameter will change like amount, gasPrice, gasLimit in the next request.

test/ost_composer/accept_stake.js
```diff
@@ -0,0 +1,182 @@
...
+//
+// ----------------------------------------------------------------------------
+//
+// http://www.simpletoken.org/
+//
+// ----------------------------------------------------------------------------
+
+const OSTComposer = artifacts.require('TestOSTComposer');
+const SpyToken = artifacts.require('SpyToken');
+const Gateway = artifacts.require('SpyEIP20Gateway');
```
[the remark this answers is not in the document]
gulshanvasnani: Added below test cases which needs SpyEIP20Gateway and SpyToken contract :-
1. should be able to transfer the base tokens to staker's staker proxy contract address.
2. should be able to transfer the value tokens to staker's staker proxy contract address.

test/ost_composer/revoke_stake_request.js
```diff
@@ -0,0 +1,201 @@
...
+      stakeRequestsStorage.staker,
+      Utils.NULL_ADDRESS,
+      'Staker address must be reset to null address',
+    );
+
+    assert.strictEqual(
+      stakeRequestsStorage.gateway,
+      Utils.NULL_ADDRESS,
+      'Gateway address must be reset to null address',
+    );
```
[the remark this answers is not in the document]
gulshanvasnani: Below test case assert the values from SpyToken :- 
`should be able to transfer the value tokens to staker`

contracts/gateway/OSTComposer.sol
schemar: I will address this after the stand-up. I am not sure.

schemar: We agreed to create a separate ticket to assert that we follow the EIP712 specification everywhere.

PR #743, opened 2019-05-16
Generate proof data

schemar:
Fixes #718

PR #751, opened 2019-05-21
Add test case that tests all generated proofs

### LINKED ISSUE #719: Write extensive unit tests for MerklePatriciaProof library
Given a valid test dataset in json format in files, write unit tests of `verify` function in the MerklePatriciaProof library.

schemar:
Tests all cases in parallel using `Promise.all`.

Fixes #719

### DIFF
--- a/test/lib/merkle_patricia_proof.js
+++ b/test/lib/merkle_patricia_proof.js
@@ -20,9 +20,13 @@
 
 const MerklePatriciaProof = artifacts.require('./MerklePatriciaProof.sol');
 const proofJson = require('../data/proof.json');
+const generatedProofData = require('../data/generatedProofData.json');
 const Utils = require('../test_lib/utils.js');
 
-contract('MerklePatriciaProof.verify()', (accounts) => {
+/** Converts a buffer that was unmarshalled into a JavaScript object into a hex string. */
+const jsonBufferToHex = json => `0x${Buffer.from(json).toString('hex')}`;
+
+contract('MerklePatriciaProof.verify()', () => {
   let merklePatriciaLib;
 
   describe('Test Cases for Account proof verification', async () => {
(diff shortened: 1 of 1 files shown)


test/lib/merkle_patricia_proof.js
```diff
@@ -37,7 +41,33 @@ contract('MerklePatriciaProof.verify()', (accounts) => {
         proofJson.account.rlpParentNodes,
         proofJson.account.stateRoot,
       );
-      assert.equal(proofStatus, true);
+      assert.strictEqual(proofStatus, true);
+
+      // Iterate over the stored data set and test all proofs in parallel.
+      // The data set was pre-generated with the fuzzy proof generator (see `./tools`).
+      await Promise.all(
+        generatedProofData.map(
```
sarvesh-ost: Just a suggestion, should this loop be moved above `it` statement and `it` statement description can also contain a pattern of proof like `ebl`. This way this will be evident what pattern has caused proof failure. Otherwise, it would be hard to track the pattern.

schemar:
Updating based on hex values from #743

test/lib/merkle_patricia_proof.js
schemar: Not sure what you mean. The "loop" is verifying each pattern with the library. How can I move that outside the `it` statement? 🤔

The pattern is actually not available here. Should we add the pattern to the proof data? It's a *little* extra effort. I don't think it's *too* much. Would also require changes in #743.

What do you think?

schemar:
Ready for review again.

/cc @sarvesh-ost

test/lib/merkle_patricia_proof.js
sarvesh-ost: I was something like this. 

```

  describe('Test Cases for Account proof verification', async () => {


    before(async () => {
      merklePatriciaLib = await MerklePatriciaProof.deployed();
    });


    it('Should pass when all the parameters are valid', async () => {
      const proofStatus = await merklePatriciaLib.verify.call(
        proofJson.account.sha3EncodedAccount,
        proofJson.account.hashedPath,
        proofJson.account.rlpParentNodes,
        proofJson.account.stateRoot,
      );
      assert.strictEqual(proofStatus, true);
      });


      generatedProofData.map(
        async (data) => {
       it(`Should pass when all the parameters are valid for the pattern ${data.pattern}` , async() => {
          const result = await merklePatriciaLib.verify.call(
            data.value,
            data.encodedPath,
            data.rlpParentNodes,
            data.root,
          );

          assert.strictEqual(
            result,
            true,
            `Could not verify data point: ${JSON.stringify(data)}`,
          );
        })
      })
  });
```

Advantage of this is, we will know for what kind of pattern the verification has failed. Otherwise, it would require manual analysis for this. 
Will it be easy to add pattern in the generate proof json?

schemar: Like I said, the pattern is unfortunately no longer available at this point.

However, `${JSON.stringify(data)}` is part of the assert message and will print exactly the data that was responsible for the failing test. Is that what you mean?

sarvesh-ost: I understand data will be part of assert message. is there any easy way to identify what pattern was used to generate this data? 
I am emphasizing on the pattern because they are very useful to reproduce the scenario. Otherwise, someone needs to manually decode `data` to know the pattern.   

Do you think the pattern is a useful data point in debugging?

sarvesh-ost: I am proposing to add pattern in the proof data 👼

schemar: #743 is already closed, so it should be a follow-up ticket ;)

schemar: I will create one

schemar: Created #754

schemar:
Ready for review and no longer blocked (rebased)

PR #755, opened 2019-05-23
Update unit test to handle arbitrary token decimals

### LINKED ISSUE #735: Update unit test to handle arbitrary token decimals.
Update unit test to handle arbitrary token decimals.
Detail of the ticket is here https://github.com/OpenST/JLP/issues/98#issuecomment-492484237

schemar:
* Added config in tests to get configured decimals (default or `env` override).
* Updated `MockToken` to take decimals as constructor argument.
* Updated all instantiations of various tokens in unit tests to use decimals from config.
* Updated proof generator to use 18 decimals (it uses mock token).
* Added new script `test:range`. Example: `npm run test:range -- 16 18`

Fixes #735

### DIFF
--- a/tools/test_range.sh
+++ b/tools/test_range.sh
@@ -28,23 +28,34 @@
 # ```bash
 # npm run test:range -- 12 18
 # ```
+#
+# You can provide only one argument and the tests will run for that specific
+# decimal.
 ###
 
 echo ""
 
-if [ $# -ne 2 ];
+if [ $# -ne 1 ] && [ $# -ne 2 ];
 then
     echo "Wrong number of parameters given."
     echo "Run with two arguments: start and end of range inclusive."
     echo "Example:"
     echo "npm run test:range -- 16 20"
     echo ""
+    echo "If you provide only one argument, the tests will be run for that value."
+    echo ""
     exit 1
 fi
 
 # Start and end values should be provided on the command line.
 START=$1
-END=$2
+
+if [ $# -eq 1 ];
+then
+    END=$START
+else
+    END=$2
+fi
 
 for DECIMALS in $(seq "${START}" "${END}");
 do
--- a/tools/test_range.sh
+++ b/tools/test_range.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Copyright 2019 OpenST Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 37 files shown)


schemar:
Ready for reviow 🐉

tools/test_range.sh
```diff
@@ -0,0 +1,58 @@
...
+# Example run with decimals 12 - 18 (inclusive):
+#
+# ```bash
+# npm run test:range -- 12 18
+# ```
+###
+
+echo ""
+
+if [ $# -ne 2 ];
```
jasonklein: The acceptability criteria for this ticket indicate that entering 1 value should be acceptable.

Although, given your `seq` approach, I think including a statement in the echo documentation that providing the same value twice will achieve that result would be a simple way to satisfy the motivation behind this particular criterion.

tools/test_range.sh
```diff
@@ -0,0 +1,58 @@
...
+then
+    echo "Wrong number of parameters given."
+    echo "Run with two arguments: start and end of range inclusive."
+    echo "Example:"
+    echo "npm run test:range -- 16 20"
```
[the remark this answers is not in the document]
schemar: I don't know if it is required, but it is the [official way](https://docs.npmjs.com/cli/run-script):
`npm run-script <command> [--silent] [-- <args>...]`

Should I remove them if it works without? I only tested with.

tools/test_range.sh
schemar: I will add that statement 👍

schemar:
> * Should `MosaicCoreConfig` be updated?

I don't understand. Can you please elaborate?

> * Should the range script end with the first value to have failing tests (current behavior) or handle failing tests with each value in the range?

I will change the script to run all the tests and change the exit code based on whether a test fails.

tools/test_range.sh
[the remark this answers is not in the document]
jasonklein: Ah! Thanks for the docs reference! I think it's better to leave the `--`. Please disregard my suggestion 🙏

jasonklein:
> > * Should `MosaicCoreConfig` be updated?
> 
> I don't understand. Can you please elaborate?

Apologies for not being specific. From `MosaicCoreConfig`:
```
/** The number of decimals of the base token. */
uint256 public constant DECIMALSFACTOR = 10 ** uint256(18);
```
I asked because of the changes to base token references in this PR. On the other hand, perhaps those references shouldn't be dynamic and should always have 18 decimals?

schemar:
> > > * Should `MosaicCoreConfig` be updated?
> > 
> > 
> > I don't understand. Can you please elaborate?
> 
> @schemar: apologies for not being specific. From `MosaicCoreConfig`:
> 
> ```
> /** The number of decimals of the base token. */
> uint256 public constant DECIMALSFACTOR = 10 ** uint256(18);
> ```
> 
> I asked because of the changes to base token references in this PR. On the other hand, perhaps those references shouldn't be dynamic and should always have 18 decimals?

Hm 🤔 Good point. Mosaic core is way outdated by now. It's probably not worth it to mingle around in the core code in the context of this PR.

tools/test_range.sh
schemar: The script will accept one and two arguments :+1

schemar:
Ready for review again. Now accepts only one argument and continues after failed decimal (prints all failed decimals at the end).

PR #757, opened 2019-05-27
Run integration and tool tests with settable decimals

### LINKED ISSUE #736: Update integration tests to handle arbitrary token decimals.
Update integration tests to handle arbitrary token decimals.
Detail of the ticket is here https://github.com/OpenST/JLP/issues/98#issuecomment-492484237

schemar:
Fixes #736

I updated the tests that test the tools as well as the integration tests
to use by default 18 decimals. The value can be overridden in all the
tests (also unit tests) with the same ENV variable (`OPENST_DECIMALS`).

The `test:range` still runs simply `npm test`, which by default runs the
`"test"` script from within `package.json`.

However, I updated how the tests are organized so that `npm test` runs
all the tests, and the individual tests have script names
`test:<section>`.

I moved "generate test proofs" out of the `test` namespace, as that is not *actually* running tests, it is just using `mocha`.


deepesh-kn:
Reviewing 🦖

Issue #772, opened 2019-06-12
Remove activeStakeRequestCount from OSTComposer contract


pgev:
`activeStakeRequestCount` is used to check if there is an active stake request for a staker across all gateways.

However, if we remove a staker proxy despite the fact if there is an active stake request, `acceptStakeRequest` would fail (as there is no staker proxy to proceed), however, an owner or a worker can `revokeStakeRequest` or `rejectStakeRequest`.

PR #775, opened 2019-06-20
Removes activeStakeRequestCount from OSTComposer

pgev:
- Updates test cases accordingly.
  - Adds `npm run update` script to ease updating submodules and npm
    packages after switching branches.

Fixes: #772

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "web3": "1.0.0-beta.36"
   },
   "scripts": {
-    "update": "git submodule update --init --recursive && npm ci",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "web3": "1.0.0-beta.36"
   },
   "scripts": {
+    "update": "git submodule update --init --recursive && npm ci",
```
gulshanvasnani: It should also run `npm ci` for the submodules. In mosaic contracts, we have added submodule for utilitytoken contracts.

pgev: Agree. Done.

PR #783, opened 2019-07-09
Release 0.11 changes

deepesh-kn:
This PR includes the changes related to npm publish.
Merging to release-0.11 branch


tsconfig.json
```diff
@@ -20,12 +20,13 @@
             "node",
             "mocha"
         ],
-        "resolveJsonModule": true
+        "resolveJsonModule": true,
+        "esModuleInterop": true,
```
pgev: I am not sure why do we need this? Do we have a module that we want to default import however it has no default export?

tools/contract_interact_generator.js
```diff
@@ -0,0 +1,89 @@
...
+if (!fs.existsSync(contractPath)) {
+  throw new Error(
+    `Cannot read file ${contractPath}.`
+    + 'Build package must run before generating contract interacts.'
+    + 'That should be done automatically when running `npm publish`.',
+  );
+}
+
+// This variable holds all the import statements as string.
+let imports = 'import { ContractOptions } from "web3-eth-contract";\nimport * as contracts from "../contract_build/contracts.json";\n';
```
pgev: Using `"` is breaking our own code style.

tools/contract_interact_generator.js
```diff
@@ -0,0 +1,89 @@
...
+    interacts = `${ interacts } 
+  /**
+   * ${contract} contract interact object
+   * @param web3 Web3 object.
+   * @param address ${contract} contract address.
+   * @param options Contract options.
+   *
+   * @returns The ${contract} contract interact object.
+   */
+  get${contract}: (web3: any, address?: string, options?: ContractOptions)=> {
```
pgev: Why do we use `any` if we are using a 55 version of web3 that has types?

pgev: What is the return type for this function?

README.md
```diff
@@ -33,28 +33,21 @@ npm install @openst/mosaic-contracts
...
-const anchorAbi = Anchor.abi;
-const anchorBinary = Anchor.bin;
+const anchorAbi = contracts.Anchor.abi;
+const anchorBinary = contracts.Anchor.bin;
+```
+
+`@openst/mosaic-contracts` also includes the contract interacts.
+
+Example of using the contract interacts. 
+```js
```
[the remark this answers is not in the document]
deepesh-kn: Made changes, added an example.

index.ts
```diff
@@ -0,0 +1,4 @@
+import * as contracts from './contract_build/contracts.json';
```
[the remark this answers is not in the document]
deepesh-kn: I have made it `contracts`, `interacts`

package.json
```diff
@@ -49,15 +49,16 @@
     "ts-node": "8.1.0",
     "typescript": "3.4.5",
     "wait-port": "0.2.2",
-    "web3": "1.0.0-beta.37",
+    "web3": "1.0.0-beta.55",
```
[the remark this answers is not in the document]
deepesh-kn: reverted to `1.0.0-beta.37`

package.json
```diff
@@ -67,13 +68,14 @@
...
     "ganache": "sh tools/runGanacheCli.sh",
     "build-package": "node tools/build_package.js",
-    "prepare": "npm run compile:all && npm run build-package",
     "deploy:gateway": "npm run compile && node tools/blue_deployment/scripts/step1_origin_contracts.js",
     "lint": "eslint {test,test_integration,tools,proof_generation} -c .eslintrc.json --ext .js --ext .ts",
-    "generate:interacts": "ts-generator ts-generator.json",
-    "postinstall": "npm run compile:all && npm run generate:interacts"
+    "generate:interacts": "ts-generator ts-generator.json && node tools/contract_interact_generator.js",
+    "copy:interacts": "cp interacts/*.d.ts dist/interacts",
+    "prepare_publish": "rm -r contract_build/contracts.json dist/* interacts/* build/* && npm run compile && npm run build-package && npm run generate:interacts && npm run compile:ts && npm run copy:interacts"
```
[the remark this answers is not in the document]
deepesh-kn: Added `npm run clean`

`npm run build-package` is called because contract interacts uses the `contract.json` which is the output of this command

package.json
```diff
@@ -67,13 +68,14 @@
     "test:unit": "truffle test",
     "ganache": "sh tools/runGanacheCli.sh",
     "build-package": "node tools/build_package.js",
-    "prepare": "npm run compile:all && npm run build-package",
     "deploy:gateway": "npm run compile && node tools/blue_deployment/scripts/step1_origin_contracts.js",
     "lint": "eslint {test,test_integration,tools,proof_generation} -c .eslintrc.json --ext .js --ext .ts",
-    "generate:interacts": "ts-generator ts-generator.json",
-    "postinstall": "npm run compile:all && npm run generate:interacts"
+    "generate:interacts": "ts-generator ts-generator.json && node tools/contract_interact_generator.js",
+    "copy:interacts": "cp interacts/*.d.ts dist/interacts",
```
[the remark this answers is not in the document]
deepesh-kn: It will create `interacts` folder

package.json
```diff
@@ -67,13 +68,14 @@
...
+    "prepare_publish": "rm -r contract_build/contracts.json dist/* interacts/* build/* && npm run compile && npm run build-package && npm run generate:interacts && npm run compile:ts && npm run copy:interacts"
   },
   "main": "dist/index.js",
```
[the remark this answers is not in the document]
deepesh-kn: This file will be generated while publishing.

tools/build_package.js
```diff
@@ -81,4 +83,5 @@ contractNames.forEach((contract) => {
   }
 });
 
-fs.writeFileSync('dist/contracts.json', JSON.stringify(contracts));
+fs.ensureDirSync('contract_build');
+fs.writeFileSync('contract_build/contracts.json', JSON.stringify(contracts));
```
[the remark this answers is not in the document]
deepesh-kn: This was the exact reason for the problem we faced, dist folder will be just the output of compiled code, it cannot overwrite at the same location.

tools/contract_interact_generator.js
```diff
@@ -0,0 +1,89 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ----------------------------------------------------------------------------
+//
+// http://www.simpletoken.org/
```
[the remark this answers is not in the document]
deepesh-kn: 😒This is consistent with all the files in the repo. We are using this all files, we will change this for all.

tsconfig.json
deepesh-kn: I read somewhere that we need to use this when `"module": "commonjs"`

tools/contract_interact_generator.js
deepesh-kn: Fixed.

tools/contract_interact_generator.js
deepesh-kn: reverted to 37 :(

deepesh-kn: Added return type

.travis.yml
```diff
@@ -24,10 +24,12 @@ notifications:
 node_js:
   - "8"
 before_script:
+  - npm run compile
+  - npm run build:package
+  - npm run generate:interacts
   - ./tools/runGanacheCli.sh </dev/null 1>/dev/null 2>&1 &
   - npm run compile:ts
```
pgev: I would propose to move this one after `npm run compile` ... kind a makes sense to me.

package.json
```diff
@@ -45,35 +46,38 @@
     "rlp": "2.2.3",
     "solidity-coverage": "0.5.11",
     "solparse": "2.2.5",
-    "truffle": "5.0.19",
+    "truffle": "beta",
```
pgev: Do we really need to do this?

package.json
```diff
@@ -45,35 +46,38 @@
...
+    "typechain": "0.3.14",
+    "fs-extra": "7.0.1"
   },
   "scripts": {
     "update": "git submodule update --init --recursive && npm ci",
     "compile": "truffle compile",
-    "compile:ts": "tsc --target es5",
-    "compile:all": "truffle compile --all && tsc --target es5",
-    "generate_test_proofs": "./proof_generation/main.sh",
+    "compile:ts": "tsc --resolveJsonModule",
```
pgev: --resolveJsonModule should not be needed in my understanding as in tsconfig.json we've set it to true.

package.json
```diff
@@ -45,35 +46,38 @@
...
     "ganache": "sh tools/runGanacheCli.sh",
-    "build-package": "node tools/build_package.js",
-    "prepare": "npm run compile:all && npm run build-package",
+    "build:package": "node tools/build_package.js",
     "deploy:gateway": "npm run compile && node tools/blue_deployment/scripts/step1_origin_contracts.js",
     "lint": "eslint {test,test_integration,tools,proof_generation} -c .eslintrc.json --ext .js --ext .ts",
-    "generate:interacts": "ts-generator ts-generator.json",
-    "postinstall": "npm run compile:all && npm run generate:interacts"
+    "generate:interacts": "ts-generator ts-generator.json && node tools/contract_interact_generator.js",
+    "clean": "rm -r contract_build/contracts.json dist/* interacts/* build/*",
```
pgev: This command will fail if contract/build/contracts.json or dist or interacts or build does not exist. Believe, this is not what we need. Hence, I would propose to do the following:

```bash
rm -r contract_build/contracts.json dist/* interacts/* build/* 2> /dev/null || true
```

tools/contract_interact_generator.js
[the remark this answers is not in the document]
pgev: Agree.

tsconfig.json
pgev: Really? What happens if we remove it?

package.json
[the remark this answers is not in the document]
pgev: I can see that, however, then we rename some commands and tool names:

npm run prepare:contracts
npm run prepare:interacts
npm run prepare:publish

We can do this in the future also.

.travis.yml
deepesh-kn: I moved `npm run compile:ts` after `npm run generate:interacts`
compile ts will compile index.ts, which imports file from autogenerated interacts, 
So this sequence is needed.

package.json
deepesh-kn: done 👍

package.json
deepesh-kn: Removed this.

package.json
deepesh-kn: Reverted this change.

tsconfig.json
deepesh-kn: Removed this, and it worked, thanks 👍

tsconfig.json
```diff
@@ -8,24 +9,21 @@
         "strictFunctionTypes": true,
         "strictPropertyInitialization": true,
         "noImplicitThis": true,
-        "noImplicitAny": true,
+        "noImplicitAny": false,
```
pgev: Let's agree and bring this back after web3 package update (I am assuming this should be the only reason).

PR #787, opened 2019-07-19
OSTComposer bug fixes

gulshanvasnani:
PR changes `valuetoken` variable name to `token` because deployed gateway contracts has `token` named variable which means value token address.
It also adds staker proxy address in `StakeRequested` event.

Reviewer can suggest better approach.

### DIFF
--- a/contracts/gateway/StakerProxy.sol
+++ b/contracts/gateway/StakerProxy.sol
@@ -23,6 +23,7 @@ pragma solidity ^0.5.0;
 import "./EIP20GatewayInterface.sol";
 import "../lib/EIP20Interface.sol";
 import "../lib/Mutex.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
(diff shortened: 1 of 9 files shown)


contracts/gateway/StakerProxy.sol
```diff
@@ -178,11 +184,21 @@ contract StakerProxy is Mutex {
...
+         * approval will override the first approval. So, in this case, total
+         * approval amount is sum of bounty and staked amount.
+         */
+        if (address(valueToken) == address(baseToken)) {
+            uint256 totalAmountToBeApproved = _amount.add(bounty);
+            valueToken.approve(address(_gateway), totalAmountToBeApproved);
+        } else {
+            valueToken.approve(address(_gateway), _amount);
+            baseToken.approve(address(_gateway), bounty);
+        }
```
deepesh-kn: Please add tests for this change.

PR #802, opened 2019-12-09
`npm run compile` should use node modules truffle version for contracts compilation

### LINKED ISSUE #801: Publish metadata of contracts while publishing package
<strike>

**Problem**

For compilation of contracts, publisher global truffle version is used. This is a problem since truffle global version can be different everytime mosaic contracts npm is published.
```
"compile": "truffle compile",
```

**Solution**

Use node modules truffle version for contracts compilation
```
"compile": "./node_modules/.bin/truffle compile",
```
</strike>

**Problem**
Contracts are being compiled on different compiler version

**Proposed Solution**
Publish metadata of contracts when publishing package

abhayks1:
**Problem**
For compilation of contracts, publisher global truffle version is used. This is a problem since truffle global version can be different everytime mosaic contracts npm is published.
```
"compile": "truffle compile",
```

**Solution**
Use node modules truffle version for contracts compilation
```
"compile": "./node_modules/.bin/truffle compile",
```

Fixes #801


pgev (changes requested):
@abhayks1 as far as I know npm run always uses the binaries first from 'node_modules/.bin', hence we do not explicitly do what is proposed in ticket.

abhayks1:
Adding the same comment which I added in ticket

@pgev Thank you for the comment. I did a quick test and it seems you are right.

Strangely contracts on etherscan are being verified on different solc versions(other than 0.5.0). Please see the report for contracts verification.
https://github.com/mosaicdao/mosaic-pm/blob/master/specification-discussions/20191206-etherscan-verification.md#chain-contracts-verification
This triggered that when publishing package, somehow compilation was done on different solc versions.

Any idea how could that happen?

abhayks1:
Closing the PR as description of ticket has changed.

